PR #1, opened 2020-04-04
Strip back repo to prepare for triage service

richardTowers:
This repo was originally mirrored from [alphagov/govuk-coronavirus-vulnerable-people-form](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form).

This PR removes all the code specific to the extremely vulnerable people service, hopefully leaving behind a skeleton that's consistent with the other applications. I also stole a tiny bit of the postgres setup from the business volunteers form (but I didn't bother stealing the redis bits).


README.md
```diff
@@ -1,6 +1,8 @@
-# CoronavirusForm
+# GOV.UK coronavirus - triage vulnerable people
```
huwd, line 1:
```suggestion
# GOV.UK coronavirus - find support service
```

I really like the name [suggested in this thread](https://gds.slack.com/archives/C010WFLGR4L/p1586160803001200?thread_ts=1586009941.002700&cid=C010WFLGR4L), could we use it for all non-repo names?

richardTowers: I think we've got time to rename the repo if we want.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: ruby
 cache: bundler
 before_script:
-- docker run -d -p 8000:8000 amazon/dynamodb-local
+- bundle exec rake db:create db:migrate
 services:
-- docker
+- postgresql
```
huwd, line 6: Do we even need postgres for this one?

Current spec is no persistance?

Gemfile
```diff
@@ -4,19 +4,18 @@ ruby File.read(".ruby-version").strip
...
-
 gem "asset_sync"
 gem "bootsnap", "~> 1"
-gem "dynamoid"
 gem "fog-aws"
 gem "govuk_app_config", "~> 2.1.1"
-gem "govuk_publishing_components", "~> 21.37.0"
+gem "govuk_publishing_components", "~> 21.38.0"
 gem "lograge"
+gem "pg", "~> 1"
```
huwd, line 13: [See above](https://github.com/alphagov/govuk-coronavirus-triage-vulnerable-people/pull/1/commits/eb1d7887954bb2748a1e2c72e4fdcf9ac4721792#r403925671) might be able to remove

.travis.yml
richardTowers: I'm afraid the conversations I had on Friday evening / Saturday morning implied that's not the case - people will want to be able to do "business intelligence" things (e.g. how many people said "Yes" to question 4). Don't see how we could do that without some data storage.

README.md
huwd: Yeah, i really like find coronavirus - find support service, so long as it doesn't confuse with the shielded service stuff?

Shall we see what the powers that be say?

richardTowers: Let's just JFDI it. I'll update the PR to use that name.

.travis.yml
huwd: yeah ok so the biz intelligence stuff... 🤔

config/locales/en.yml
```diff
@@ -128,192 +103,3 @@ en:
       <p class="govuk-body">Your session has ended because you have not done anything for 4 hours. You'll have to start again.</p>
 
       <p class="govuk-body">We do this for your security. We've deleted all the details you entered to protect your data.</p>
-  coronavirus_form:
```
leenagupte, line 131: I wonder if this heading, the "errors" section and an empty "questions" section should be kept in?

features/form.feature
```diff
@@ -1,61 +1,5 @@
...
-    When I answer "No"
-    Then I can see "Are you ready to send your application?" in the heading
-    When I click the "Accept and send" button
-    Then I can see "Registration complete"
-
-  Scenario: Visits an intermediate question without having previously visited the form
-    When I visit the "/essential-supplies" path
-    Then I will be redirected to the "/live-in-england" path
-
-  Scenario: Visits the privacy policy
```
leenagupte, line 43: I think we can keep this test.

richardTowers: Oh yeah - good point!

richardTowers: Ah, we can't - the first thing it does is got to /live-in-england, and we don't have an equivalent of that yet because I missed all the questions.

Could make it simpler and just go to `/privacy`, but that involves changing the test so let's do it in a separate PR I think.

config/locales/en.yml
```diff
@@ -129,191 +104,5 @@ en:
 
       <p class="govuk-body">We do this for your security. We've deleted all the details you entered to protect your data.</p>
   coronavirus_form:
-    errors:
-      page_title_prefix: 'Error: '
-      heading: "There is a problem"
```
leenagupte, line 134: Is it worth keeping these two? They seem pretty standard.

features/form.feature
leenagupte: Were you going to add this back in?

richardTowers: I can't without rewriting the test - it tries to go to /live-in-england

config/locales/en.yml
richardTowers: Let's ship it, and we can always add them back.

PR #8, opened 2020-04-07
Add "are you still working" question

brucebolt:
Adds controller, view, route and request test for this question.

I've implemented a request test here since that allows us to test the behaviour of the controller and view at the same time, covering behaviour that is untested in the other two form applications.

Next question path, previous question path and test for redirecting cannot yet be implemented (and therefore I've added placeholders) since those questions are not yet available.

Preview:
Screenshot 2020-04-07 at 15 27 33

Trello card: https://trello.com/c/SdBbChS1


brucebolt:
We don't yet have any content for the validation errors, so I'll merge this anyway and we can update once we have content.

PR #10, opened 2020-04-07
Add heading subtitle to still working question

injms:
The content doc was updated, which meant that the need for a heading subtitle was made clearer. 

This adds it to the question already done 'cos @brucebolt is just super speedy.

## Visual differences

Before:
image

---

After:
image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -108,6 +108,7 @@ en:
     questions:
       still_working:
         title: "Are you still going in to work even though you’re not a key worker?"
+        title_caption: "Going in to work"
         options:
           - "Yes"
           - "I do not know if I’m a key worker"
--- a/app/views/coronavirus_form/still_working.html.erb
+++ b/app/views/coronavirus_form/still_working.html.erb
@@ -16,6 +16,7 @@
 
 <%= render "govuk_publishing_components/components/radio", {
   heading: t("coronavirus_form.questions.still_working.title"),
+  heading_caption: t("coronavirus_form.questions.still_working.title_caption"),
   is_page_heading: true,
   name: "still_working",
   error_message: error_items("still_working"),


config/locales/en.yml
```diff
@@ -108,6 +108,7 @@ en:
     questions:
       still_working:
         title: "Are you still going in to work even though you’re not a key worker?"
+        title_caption: "Going in to work"
```
leenagupte, line 111: I wonder whether the structure of the YAML should be re-jigged a bit because quite a few questions are going to need this caption?

Perhaps:
```
coronavirus_form:
  groups:
      going_to_work:
         title: Going in to work
         questions:
             still_working:  
                  title: Are you still going in to work even though you’re not a key worker?
...
```

Then you'd be able to reference the same group across different views:
`t("coronavirus_form.groups.going_to_work.title")`

But the rest of the references would need to be expanded to include `groups.going_to_work`

What do you think?

injms: I think this is a good structure for the question that have subtitles. There are three questions without subtitles we'd need to invent groups for - not a big deal, but another thing to consider.

Currently only one of the groups has three questions in it; three groups have two questions, and the remaining six groups have a single question. You'll end up with one subtitle repeated three times, and another subtitle repeated twice. I don't think this is duplication is a problem in a data structure that people are already familiar with.

I think that the structure is neat and a good idea, but think for this case this is premature optimisation. But I do think we should revisit this later if the i18n YAML becomes too repetitive.

leenagupte: Assuming that the design of the results page doesn't change, know which group a question belongs to will useful for that too.

huwd: Makes sense to me!

leenagupte: I think we were properly modelling each question, we could get away with having a “group” property on the question. Something for @huwd and I to think about.

injms:
I'm closing this as it's been fixed by #29.

PR #12, opened 2020-04-08
Add "Worried about work because living with vulnerable" question

JonathanHallam:
Adds controller, view, route and request test for this question.

### DIFF
--- a/app/controllers/coronavirus_form/living_with_vulnerable_controller.rb
+++ b/app/controllers/coronavirus_form/living_with_vulnerable_controller.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::LivingWithVulnerableController < ApplicationController
+  def submit
+    @form_responses = {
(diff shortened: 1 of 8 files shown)


app/controllers/coronavirus_form/living_with_vulnerable_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    invalid_fields = validate_radio_field(
+      controller_name,
+      radio: @form_responses[:living_with_vulnerable],
+    )
+
+    if invalid_fields.any?
+      flash.now[:validation] = invalid_fields
+      log_validation_error(invalid_fields)
+      render controller_path
+    elsif session["check_answers_seen"]
```
injms, line 18: I don't think we need this `elsif`, since we don't have a check answers page.

JonathanHallam: 👍

PR #13, opened 2020-04-08
Add "Are you off work because you’re ill...?" question

injms:
Adds the controller, view, routes, content, and request test for the "Are you off work because you’re ill...?" question.

[Trello card](https://trello.com/c/ntCuSOGF)

image

image

### DIFF
--- a/app/controllers/coronavirus_form/are_you_off_work_ill_controller.rb
+++ b/app/controllers/coronavirus_form/are_you_off_work_ill_controller.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::AreYouOffWorkIllController < ApplicationController
+  def submit
+    @form_responses = {
+      are_you_off_work_ill: strip_tags(params[:are_you_off_work_ill]).presence,
+    }
+
+    invalid_fields = validate_radio_field(
+      controller_name,
+      radio: @form_responses[:are_you_off_work_ill],
+    )
+
+    if invalid_fields.any?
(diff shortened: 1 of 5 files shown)


app/controllers/coronavirus_form/are_you_off_work_ill_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::AreYouOffWorkIllController < ApplicationController
+  skip_before_action :check_first_question
```
brucebolt, line 4: This skip check is only needed for the first question in the application

config/locales/en.yml
```diff
@@ -106,6 +106,13 @@ en:
   coronavirus_form:
     errors: {}
     questions:
+      are_you_off_work_ill:
+        title: "Are you off work because you’re ill or self-isolating?"
+        title_caption: "Going in to work"
+        options:
+          - "Yes"
+          - "No"
+        custom_select_error: "Select yes if you're off work because you're ill or self-isolating"
```
brucebolt, line 115: Tiny, pedantic point
```suggestion
        custom_select_error: "Select yes if you’re off work because you’re ill or self-isolating"
```

app/controllers/coronavirus_form/are_you_off_work_ill_controller.rb
injms: Removed 👍

PR #21, opened 2020-04-08
Add "are you able to get food" question

brucebolt:
Adds controller, view, routes and test for the "Are you able to get food?" question.

Screenshot 2020-04-08 at 10 48 04

Trello card: https://trello.com/c/xHCZ0fTX

### DIFF
--- a/spec/requests/get_food_spec.rb
+++ b/spec/requests/get_food_spec.rb
@@ -1,0 +1,52 @@
+RSpec.describe "get-food" do
+  describe "GET /get-food" do
+    let(:selected_option) { I18n.t("coronavirus_form.questions.get_food.options").sample }
+
+    context "without session data" do
+      it "shows the form" do
+        visit get_food_path
+
+        expect(page.body).to have_content(I18n.t("coronavirus_form.questions.get_food.title"))
+        I18n.t("coronavirus_form.questions.get_food.options").each do |option|
+          expect(page.body).to have_content(option)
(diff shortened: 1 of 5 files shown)


spec/requests/get_food_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      post get_food_path, params: { get_food: selected_option }
+
+      expect(response).to redirect_to(next_question_path)
+    end
+
+    xit "shows an error when no radio button selected" do
+      post get_food_path
+
+      expect(response.body).to have_content(I18n.t("coronavirus_form.questions.get_food.title"))
+      expect(response.body).to have_content(I18n.t("coronavirus_form.questions.get_food.custom_select_error"))
```
leenagupte, line 49: This test could be implemented if you use: `expect(response.body).to have_content(I18n.t("coronavirus_form.errors.radio_field"))`

brucebolt: I'm working on the basis we'll probably get custom error text for these questions, since "Select are you able to get food?" doesn't really sound right to me.

leenagupte: But that's what's being displayed at the moment if you don't select anything.

brucebolt: That's true.  Either way the test will need changing.  I'll update to use the error that's currently being shown.

PR #24, opened 2020-04-08
Add "Do you have somewhere to live?" question

JonathanHallam:
Add controller, tests and view for that question

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -46,6 +46,10 @@ Rails.application.routes.draw do
     get "/mental-health-worries", to: "mental_health_worries#show"
     post "/mental-health-worries", to: "mental_health_worries#submit"
 
+    # Question: Have you got somewhere to live?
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -46,6 +46,10 @@
     get "/have-you-been-evicted", to: "have_you_been_evicted#show"
     post "/have-you-been-evicted", to: "have_you_been_evicted#submit"
 
+    # Question: Have you got somewhere to live?
```
brucebolt, line 49: I'd suggest moving this up, so the questions remain in the same order as in the Figma prototype (https://www.figma.com/file/tibZIxmeYW0Z3qEL4CEF72/Triage-tool---prototypes?node-id=953%3A2)

issyl0:
Also needs a rebase.

config/routes.rb
JonathanHallam: Done, most of these seem to be out of order so I'll raise a PR after this to change them round

PR #25, opened 2020-04-08
Add 'Are you finding it hard to afford rent, your mortgage or bills?' question

Rosa-Fox:
affording-rent-mortgage-bills

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -117,6 +117,12 @@ en:
           - "Yes"
           - "No"
         custom_select_error: "Select yes if you need urgent medical help"
+      afford_rent_mortgage_bills:
+        title: "Are you finding it hard to afford rent, your mortgage or bills?"
+        options:
+          - "Yes"
+          - "No"
+          - "Not sure"
       get_food:
         title: "Are you able to get food?"
         title_caption: "Getting food"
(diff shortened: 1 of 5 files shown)


Rosa-Fox:
going to change `affording` to `afford` to align with `afford-food`

config/routes.rb
```diff
@@ -30,7 +30,7 @@
     get "/still-working", to: "still_working#show"
     post "/still-working", to: "still_working#submit"
 
-    # Question: Are you worried about going into work because you live with someone vulnerable to coronavirus?
+    # Question: Are you worried about going into work because you live with someone vulnerable
```
brucebolt, line 33: I'm not sure this was intentional?

config/locales/en.yml
```diff
@@ -157,6 +157,12 @@ en:
           - "Yes"
           - "No"
           - "Not sure"
+      afford_rent_mortgage_bills:
```
brucebolt, line 160: I'd suggest moving this up to line 120, so the questions remain in the same order as the Figma prototype (https://www.figma.com/file/tibZIxmeYW0Z3qEL4CEF72/Triage-tool---prototypes?node-id=953%3A2).

config/routes.rb
Rosa-Fox: ah no, error when rebasing! Thanks, have amended.

config/locales/en.yml
Rosa-Fox: Thanks! Have amended.

PR #26, opened 2020-04-08
Questions as data

huwd:
Trello: https://trello.com/c/LNk85mPk/178-represent-questions-as-data

## What/Why

This PR proposes a data structure for describing corona virus questions and content. 

The three existing corona forms all hold data about questions as code, largely in controllers. This has two downsides:
- You need a dev to dive into the code to make any edits
- There is a lot of repetition between the apps

The aim here is to come up with a way of abstracting content and question structure out into data that can then be consumed by routes, controllers and views to dynamically create, structure and order new questions whenever you add a new element in data.

Aims are:
- To come up with something easy for devs to maintain, adding a ton of unnecessary complexity won't help anyone long term. It should reflect learning from products like the Brexit checker (devs from there, please ruthlessly critique this PR!)
- To ensure that existing content can be updated by content designers through github, and that the YAML they work with is clear enough for them to use easily.
- The abstraction need only to work currently for the simple use case of this 'find-support' app. But should keep one eye on the the 'vulnerable-people' and 'biz-volunteering' apps, anything that would stop those structures being moved into this data format would be good to consider here (we can debate about what is premature optimization!) 

## Ways this can be extended
The Brexit checker offers some ideas about adding in a criteria key, so we can concatenate a question and an option key to produce a unique response to a prior question. This could then be used to show/skip a subsequent question.

## Things I'm not sure about
Routing, should we describe paths in data too and do some dynamic magic in the routes.rb file (keeps everything in data, other form builders appear to have a 'path' param)... or should we define those explicitly in the routes folder (expected rails practice?)

## Summary of structure
So i figure you can define a form structure as it's own YAML file.
That contains, an ordered array of question objects, which looks something like:
```
  key: [unique human readable key] (Required)
  group: [unique human readable group key] (Optional)
  type: [Enum of supported types eg: radio / checkbox / input / text area) (Required)
  options: Depends on type, but for checkbox / radio list options (conditionally required)
```

Could then extend that to include:
```
criteira: [object that contains rules on when to skip/show the question] (Optional)
validation: [a set of rules, eg required, or names of ruby validators that must be used with this question, otherwise it defaults to skippable] (Optional)
```

Could also define an options criteria that can signify a requirement to provide additional items to any view component (eg, hints, descriptions - [see other examples here](https://github.com/alphagov/finder-frontend/blob/master/lib/brexit_checker/questions.yaml). However i still believe the actual content for those should then be looked up in the locale file.

## Next steps
Assuming this PR makes sense:
- Data integrity checks
- A single controller for questions
- A single view for questions
- Appropriate partials.


config/locales/forms/find_support.en.yml
```diff
@@ -0,0 +1,59 @@
+en:
+  find_support:
+    defaults:
+      options:
+        yes: "Yes"
+        soon: "I might be soon"
```
leenagupte, line 6: I don't think we should have default options. I think it could get quite messy making sure options are excluded etc.

config/locales/forms/find_support.en.yml
```diff
@@ -0,0 +1,59 @@
...
+      going_in_to_work: "Going in to work"
+      feeling_unsafe: "Feeling unsafe"
+      housing: "Having somewhere to live"
+      mental_health: "Mental health and wellbeing"
+    questions:
+      are_you_out_of_employment:
+        title: "Have you been made unemployed or put on temporary leave (furloughed)?"
+      are_you_non_keyworker_still_going_to_work:
+        title: "Are you still going in to work even though you’re not a key worker?"
+        special_options:
```
leenagupte, line 24: As mentioned above, to keep it easier to read and update, I think each question should define all of it's options.

lib/forms/find_support.yml
```diff
@@ -0,0 +1,113 @@
+_
+find_support:
+- key: are_you_out_of_employment
+  group: not_in_work
+  type: radio
+  required: true
+  options:
+    _ "yes"
```
leenagupte, line 8: Are this options supposed to match the I18n options? Do the options also need to be defined here?

lib/forms/find_support.yml
```diff
@@ -0,0 +1,113 @@
...
+    _ "soon"
+    _ "no"
+    _ "not_sure"
+- key: are_you_non_keyworker_still_going_to_work
+  group: going_in_to_work
+  type: radio
+  required: true
```
leenagupte, line 15: Could you give an example of a question that can be skipped? For example, when would it be skipped?

config/locales/forms/find_support.en.yml
huwd: Yeah up for a debate on this one, my thinking was many of these forms (especially things like radio boxes), re-use the same answers over and over and over. I don't want to DRY-for-DRY sake, but i figure not having these typed out each time helps keep the YAML file shorter and slimmer, which might help content folks navigate it on github?

The mental model would be:
- each question has it's own options, anything added there overrides any other locale
- anything shared can be DRYed out to shared defaults at the top. Also means we can could change all "Yes" answers to "Yep" in one place if we wanted to.

lib/forms/find_support.yml
huwd: Yeah, i was thinking these should be the equivalent to the `value` attribute on an HTML `<input>`. So the answer is completely abstracted from the content, but because checkboxes and radios are always a known list, we can use the values to then look up the appropriate text for each answer in the locale file.

lib/forms/find_support.yml
huwd: We could invert the default value if that made sense, so it defaults to required if not explicitly told otherwise. But i think it might make sense to offer the change that a question has no validation on it.

I think the design system is also opinionated about skippable questions, we had a conversation on the Brexit checker about "Continue" buttons over "Next" etc. So we might need to allow the partials to add additional elements if things are skippable?

We see that [we did that here](https://www.gov.uk/transition-check/questions) - even if we might regret it somewhat, I can imagine there being calls for it.

config/locales/forms/find_support.en.yml
issyl0: I'm not a fan of this. I think each question should have the full amount of answers defined with the question content (like we did for the other forms). If content designers are editing the file, they often want to edit one question. Having to scroll to the top to edit "Yes" to "Yep" and not realising that that'll change for everything is potentially disastrous.

lib/forms/find_support.yml
leenagupte: Ah! I assumed "skippable" means that the user is never shown the question, because they answered "no" to something else, not that they're allowed to not answer the question.

config/locales/forms/find_support.en.yml
huwd: Humm, I suppose I'm wondering if the locale file enumerates all possibilities in a non-dry way, is that then confusing when comparing a questions data configuration and a questions content?

They have to match 100% instead of just describing where you can find content for a question?

huwd: I take the point about the potential for mix up, you still think that's a risk given the github requirement for review and the idea we could use tests to cover really unexpected outcomes?

lib/forms/find_support.yml
huwd: ah right, I've been thinking of that as "criteria", so the NHS number question is dependent on you having answered Yes to do the "do you know your NHS number question". I've parked that in the to-do-later stuff, because i don't think we have any questions with dependent criteria here (we'll revisit that checkbox perhaps).

So skipptable i guess means unvalidated, you can click next / continue and not provide a response.

config/locales/forms/find_support.en.yml
leenagupte: I'm with Issy on this. The potential for accidentally changing questions you didn't mean to is very high. We also have to account for questions that don't use any of the default options.

huwd:
Just read this:
https://guides.rubyonrails.org/i18n.html#organization-of-locale-files

I'm thinking about structure. Wondering if this is better:
```
|-locales
|--forms
|---find_support
|----answers
|-----en.yml
|----groups
|-----en.yml
|----questions
|-----en.yml
```

🤔 it's quite deeply nested but might be a more organised structure, and matches expectations for translations a bit better.

config/locales/forms/find_support.en.yml
huwd: Is that the checkbox question?

huwd: I suppose the only other argument I can think of for defaults, is that if you added another radio box question, which used the same keys as other questions but forgot to add the locale file, it'd render something sensible instead of the locale error pattern.

But potentially we can use data validation tests/checks to protect against that? 🤔 

But fair if there's a strong sentiment against it, happy to change. Will hang on for another couple reviews then switch it over to the majority opinion / the consensus point?

config/locales/forms/find_support.en.yml
huwd: main discussion on [this thread](https://github.com/alphagov/govuk-coronavirus-find-support/pull/26#discussion_r405456524)

config/locales/forms/find_support.en.yml
huwd: ok questions...
What if... i renamed the key to `fallbacks` and put it at the bottom of the file?
Is the problem that there's a risk in changes being made without realising the scope? or,
is it more about not enumerating all possibilities in the locale file?

trying to think this one through 🤔

erino:
@huwd Good idea. Sorry I've not ready this thoroughly, but I want to comment on this point

> The abstraction need only to work currently for the simple use case of this 'find-support' app. But should keep one eye on the the 'vulnerable-people' and 'biz-volunteering' apps, anything that would stop those structures being moved into this data format would be good to consider here (we can debate about what is premature optimization!)

I think there's value in looking at a solution that'll work for all of the apps. Building each app in a slightly different way will make maintenance much harder (we have this problem in GOV.UK with publishing apps built in slightly different ways). Taking slightly longer now for a more complete solution will pay off in the long run.

config/locales/en.yml
```diff
@@ -1,203 +1,204 @@
...
-          - "No"
-          - "Not sure"
+        <p class="govuk-body">We do this for your security. We've deleted all the details you entered to protect your data.</p>
+    get_help_from_nhs:
+      title: Get urgent help from the NHS now
+      link_text: Go to NHS 111 online
+      link_href: "https://111.nhs.uk/"
+    coronavirus_form:
+      errors:
+        page_title_prefix: 'Error: '
```
vanitabarrett, line 113: I realise this wasn't introduced in this PR, but if the idea is that these files are more easily editable by others, should we remove the need to specify a space in this yaml and move it into the code instead?

config/locales/en.yml
```diff
@@ -1,203 +1,204 @@
...
+        heading: "There is a problem"
+        radio_field: "Select %{field}"
+        checkbox_field: "Select at least one %{field}"
+      questions:
```
vanitabarrett, line 117: From someone completely new to this app and it's structure, I'm not sure why `questions` here is in `en.yml` and not in `questions/en.yml`.

lib/forms/find_support.yml
```diff
@@ -0,0 +1,113 @@
+_
+find_support:
+- key: are_you_out_of_employment
+  group: not_in_work
+  type: radio
+  required: true
+  options:
```
vanitabarrett, line 7: There seems to be quite a lot of repetition here, especially in defining all the options again. It seems like there's more chance of human error? 

There also seems to be an option defined above (yes_but_someone_else) which doesn't appear in this file - is that correct?

config/locales/en.yml
huwd: That's a really good point, should be moved
I think we thought we'd start out with a blob and then break it out as needed.
... i think it's needed :D

huwd:
Gonna close this, as I'm leaving the team today and some context might go with it.
TLDR on this, if we're going to build more form services like this for corona or otherwies, might be worth looking at the tooling to help churn them out quicker.

Having a way to represent not just the contnet, but also the structure as data, would help make that process easier

PR #27, opened 2020-04-08
Add checkbox question for need help with

leenagupte:
Trello: https://trello.com/c/sywMAZQz

Screenshot 2020-04-08 at 12 09 31
Screenshot 2020-04-08 at 14 23 23

### DIFF
--- a/app/controllers/coronavirus_form/need_help_with_controller.rb
+++ b/app/controllers/coronavirus_form/need_help_with_controller.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::NeedHelpWithController < ApplicationController
+  def submit
+    @form_responses = {
+      need_help_with: Array(params[:need_help_with]).map { |item| strip_tags(item).presence }.compact,
+    }
+
+    invalid_fields = validate_checkbox_field(
+      controller_name,
+      values: @form_responses[:need_help_with],
+      allowed_values: I18n.t("coronavirus_form.questions.#{controller_name}.options"),
+    )
+
+    if invalid_fields.any?
+      flash.now[:validation] = invalid_fields
+      log_validation_error(invalid_fields)
+      render controller_path
+    else
+      update_session_store
+      # redirect_to nil
+    end
+  end
+
+private
+
+  def update_session_store
+    session[:need_help_with] = @form_responses[:need_help_with]
+  end
+
+  def previous_path
+    "/"
+  end
+end
--- a/app/controllers/coronavirus_form/need_help_with_controller.rb
+++ b/app/controllers/coronavirus_form/need_help_with_controller.rb
@@ -9,7 +9,7 @@ class CoronavirusForm::NeedHelpWithController < ApplicationController
     invalid_fields = validate_checkbox_field(
       controller_name,
       values: @form_responses[:need_help_with],
-      allowed_values: I18n.t("coronavirus_form.groups.#{group}.questions.#{controller_name}.options"),
+      allowed_values: group_titles + extra_options,
     )
 
     if invalid_fields.any?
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -201,3 +201,14 @@ en:
...
+        title: "What do you need to find help with?"
+        options:
+          - "Feeling unsafe"
+          - "Paying bills"
+          - "Getting food"
+          - "Being unemployed or not having any work"
+          - "Going in to work"
+          - "Having somewhere to live"
+          - "Mental health and wellbeing"
+          - "I'm not sure"
```
issyl0, line 214:
```suggestion
          - "I’m not sure"
```

app/views/coronavirus_form/need_help_with.html.erb
```diff
@@ -0,0 +1,33 @@
...
+  "data-question-key": "need_help_with",
+  "id": "need_help_with",
+  "novalidate": "true"
+) do %>
+
+<%= render "govuk_publishing_components/components/checkboxes", {
+  heading: t("coronavirus_form.questions.need_help_with.title"),
+  is_page_heading: true,
+  name: "need_help_with[]",
+  error_message: error_items("need_help_with"),
```
alex-ju, line 21: Unfortunately, checkboxes API is inconsistent with the radios API, so we'll need to use `error` to show the error message at the component-level.
```suggestion
  error: error_items("need_help_with"),
```

config/locales/en.yml
injms, line 214: The wording need tweaking, but is it worth highlighting the 'Not sure' option in the error message?

```suggestion
          - "I'm not sure"
        custom_select_error: 'Please choose at least one option. If you don't know, please choose "Not sure"'
```

issyl0: Typographical quotes for apostrophes at all times please.

leenagupte:
@alex-ju @issyl0 @injms Thanks for your comments. I've updated and fixed up.

app/controllers/coronavirus_form/need_help_with_controller.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::NeedHelpWithController < ApplicationController
+  def submit
+    @form_responses = {
+      need_help_with: Array(params[:need_help_with]).map { |item| strip_tags(item).presence }.compact,
```
brucebolt, line 6: It would be super useful if we could store the key of the question groups that we should display to the user based on their answer to this question

leenagupte: In @form_responses or in the session?

brucebolt: In the session

leenagupte: So like:
```
session[:feeling_safe] = true
session[:paying_bills] = true
```
If they selected "Feeling safe" and "Paying bills"?
```

Comment (brucebolt) [reply app/controllers/coronavirus_form/need_help_with_controller.rb:6]: Yes, but maybe all under one session key?

E.g.
```
session[:groups_selected] = {
  felling_safe: true,
  paying_bills: true,
}
```

This will make it easier to work out which questions we should show the user.

Comment (leenagupte) [reply app/controllers/coronavirus_form/need_help_with_controller.rb:6]: Updated and pushed up

Comment (brucebolt) [reply app/controllers/coronavirus_form/need_help_with_controller.rb:6]: Thanks, this will make my current work so much easier

File: spec/requests/need_help_with_spec.rb [hunk 4]
```diff
@@ -0,0 +1,55 @@
...
+        visit need_help_with_path
+
+        expect(page.body).to have_content(I18n.t("coronavirus_form.groups.filter_questions.questions.need_help_with.title"))
+        expect(page.find("input#option_#{selected.first.parameterize.underscore}")).to be_checked
+      end
+    end
+  end
+
+  describe "POST /need-help-with" do
+    let(:selected) { ["Feeling unsafe"] }
```
huwd, line 31: Just gave these a test, I can't get it to submit.

I think it's because there's a gap between the submitted value which would be `feeling_unsafe` and the tests + validations expectations that it would be the Intl Value `Feeling unsafe`

app/controllers/coronavirus_form/need_help_with_controller.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::NeedHelpWithController < ApplicationController
```
huwd, line 3: Humm, I'm thinking if there's a gap between the values and human readable strings here.

Is it worth making this less dynamic and just putting it all in a constant at the top of the file?
like:
```
  QUESTION_GROUPS = {
    "feeling_unsafe": "Feeling unsafe",
    "paying_bills": "Paying bills",
    "getting_food": "Getting food",
    "being_unemployed": "Being unemployed or not having any work",
    "going_in_to_work": "Going in to work",
    "somewhere_to_live": "Having somewhere to live",
    "mental_health": "Mental health and wellbeing",
    "I’m not sure": "I’m not sure",
  }.freeze
```

As far as I can tell, those are the keys coming in from the form, the I'm not sure key is a bit odd, but i think that's because the key is being taken from the locale instead, so is inconsistent with the others?

leenagupte: It worked when I was plucking the values from the I18n file.

leenagupte: I'm going to see if I can figure it out with a fresher brain.

richardTowers:
"Select at least one what do you need help with?" doesn't feel like a great error message (from your screenshot - I haven't actually tested). Not sure whether that needs fixing in this PR, or if we want to wait for copy for the error messages from the content designers?

app/controllers/coronavirus_form/need_help_with_controller.rb
leenagupte: @huwd I think I figured it out. Code is super ugly though.

leenagupte:
> ~"Select at least one what do you need help with?" doesn't feel like a great error message (from your screenshot - I haven't actually tested). Not sure whether that needs fixing in this PR, or if we want to wait for copy for the error messages from the content designers?~
> 
> (edit: looking at the code, I think you've already solved this. Nice one!)

@richardTowers I've updated the screenshot 🎉

PR #29, opened 2020-04-08
Update i18n structure

injms:
This now puts the questions in groups, which should make it easier to pull out what needs to be shown for the results page.

### DIFF
--- a/app/controllers/coronavirus_form/afford_food_controller.rb
+++ b/app/controllers/coronavirus_form/afford_food_controller.rb
@@ -8,7 +8,7 @@ class CoronavirusForm::AffordFoodController < ApplicationController
 
     invalid_fields = validate_radio_field(
       controller_name,
-      "getting_food",
+      group,
       radio: @form_responses[:afford_food],
     )
 
@@ -30,5 +30,9 @@ private
 
   def previous_path
     "/"
+  end
+
+  def group
+    "getting_food"
   end
 end
--- a/app/controllers/coronavirus_form/afford_food_controller.rb
+++ b/app/controllers/coronavirus_form/afford_food_controller.rb
@@ -8,6 +8,7 @@ class CoronavirusForm::AffordFoodController < ApplicationController
 
     invalid_fields = validate_radio_field(
       controller_name,
+      "getting_food",
       radio: @form_responses[:afford_food],
     )
 
--- a/app/views/coronavirus_form/have_somewhere_to_live.html.erb
+++ b/app/views/coronavirus_form/have_somewhere_to_live.html.erb
@@ -1,1 +1,4 @@
-<%= render partial: "question_radio_buttons", locals: {question: "have_somewhere_to_live"} %>
+<%= render partial: "question_radio_buttons", locals: {
+  question: controller_name,
+  group: group,
+} %>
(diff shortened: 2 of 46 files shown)


config/locales/en.yml
```diff
@@ -113,91 +113,117 @@ en:
...
-    questions:
-      urgent_medical_help:
-        title: "Do you need urgent medical help?"
-        description: "You can check if you have coronavirus (COVID-19) symptoms and get medical advice"
-        options:
-          - "Yes"
-          - "No"
+    groups:
+      help: # NHS question and 'What do you need help with'.
+        title: false # No title needed for this group.
```
leenagupte, line 118: This there a risk that "false" could be displayed somewhere? Should this attribute be removed instead?

app/views/development/development.html.erb
```diff
@@ -0,0 +1,19 @@
+<% content_for :title do %>Routes<% end %>
```
leenagupte, line 1: Where did this come from? It's a nice idea, but's it's another thing that needs to be kept up to date 😞

app/views/coronavirus_form/living_with_vulnerable.html.erb
```diff
@@ -1 +1,4 @@
-<%= render partial: "question_radio_buttons", locals: {question: "living_with_vulnerable"} %>
+<%= render partial: "question_radio_buttons", locals: {
+  question: "living_with_vulnerable",
+  group: "going_in_to_work",
```
Rosa-Fox, line 3: this should be ` living_with_vulnerable`

config/locales/en.yml
injms: Fixed - I've removed the key and made the default translation, which is used if there isn't a translation available, `false`.

app/views/development/development.html.erb
injms: Argh, that was me seeing if it could be done automagically. Will delete!

app/views/coronavirus_form/living_with_vulnerable.html.erb
injms: The question or the group?

Rosa-Fox: the group

Rosa-Fox: ooh sorry I am wrong, I misread

app/controllers/coronavirus_form/afford_food_controller.rb
```diff
@@ -8,6 +8,7 @@ def submit
 
     invalid_fields = validate_radio_field(
       controller_name,
+      "getting_food",
```
leenagupte, line 11: Do you think it's worth having a "group" method, a bit like "previous_path" which could be used to define the group per controller?  I think that would also allow the radio button partial to be called with:

```
<%= render partial: "question_radio_buttons", locals: {
  question: "afford_food",
  group: group,
} %>
```

leenagupte: We could probably do the same for "question"

leenagupte: It can be done!

app/views/coronavirus_form/have_somewhere_to_live.html.erb
```diff
@@ -1 +1,4 @@
-<%= render partial: "question_radio_buttons", locals: {question: "have_somewhere_to_live"} %>
+<%= render partial: "question_radio_buttons", locals: {
+  question: controller_name,
+  group: group,
```
leenagupte, line 3: Where did you come from? I think the rebase went a bit rogue.

injms: This is from #30 - updated to use the same variables as the other questions.

PR #34, opened 2020-04-08
Add escape link component

alex-ju:
- [x]  add escape link component
- [x]  use the component on question pages
- [x] cross-device testing
- [x] cross-browser testing (aware of the limited support for `sticky`, but has a decent fallback)

## Escape link component
Screenshot 2020-04-08 at 18 28 34

Screenshot 2020-04-08 at 18 28 49

## Example of usage on 'Dow you feel safe where you live' page
<table>
<tr><th>Mobile</th><th>Desktop</th></tr>
<tr><td valign="top">

localhost_5000_feel-safe

</td><td valign="top">

localhost_5000_feel-safe (1)

</td></tr>
</table>

Known issue:
If users are going back in the history and then press the link it doesn't work as it's currently not possible to detect the current `history.index` only the `history.length`. Further investigation is required to detect this state.

[Trello card](https://trello.com/c/c7UahJby)

### DIFF
--- a/app/assets/stylesheets/components/_escape-link.scss
+++ b/app/assets/stylesheets/components/_escape-link.scss
@@ -1,0 +1,27 @@
+.app-c-escape-link {
+  background-color: govuk-colour('white');
+  border-top: 1px solid $govuk-border-colour;
+  bottom: 0;
+  position: fixed;
+  position: sticky; // scss-lint:disable DuplicateProperty
+  text-align: center;
+  width: 100%;
+
+  .govuk-link {
+    @include govuk-font($size: 19, $weight: bold);
+    color: $govuk-error-colour;
+    display: block;
+    padding: govuk-spacing(6) 0;
+
+    &:link,
+    &:visited,
+    &:hover,
+    &:active {
+      color: darken($govuk-error-colour, 5%);
+    }
+
+    &:focus {
+      color: $govuk-focus-text-colour;
+    }
+  }
+}
--- a/app/views/application/_question_radio_buttons.html.erb
+++ b/app/views/application/_question_radio_buttons.html.erb
@@ -13,22 +13,27 @@
   "id": question,
   "novalidate": "true"
 ) do %>
+  <%= render "govuk_publishing_components/components/radio", {
+    heading: t("coronavirus_form.groups.#{group}.questions.#{question}.title"),
+    heading_caption: t("coronavirus_form.groups.#{group}.title", { default: false }),
+    is_page_heading: true,
+    name: question,
+    error_message: error_items(question),
+    items: t("coronavirus_form.groups.#{group}.questions.#{question}.options").map do |option|
+        {
+          value: option,
+          text: option,
+          id: "option_#{option.parameterize.underscore}",
+          checked: @form_responses[question.to_sym] == option,
+        }
+      end
+  } %>
+  <%= render "govuk_publishing_components/components/button", text: "Continue", margin_bottom: true %>
+<% end %>
 
-<%= render "govuk_publishing_components/components/radio", {
-  heading: t("coronavirus_form.groups.#{group}.questions.#{question}.title"),
-  heading_caption: t("coronavirus_form.groups.#{group}.title", { default: false }),
-  is_page_heading: true,
-  name: question,
-  error_message: error_items(question),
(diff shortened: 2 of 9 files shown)


app/views/application/_question_radio_buttons.html.erb
```diff
@@ -13,22 +13,27 @@
...
-        checked: @form_responses[question.to_sym] == option,
-      }
-    end
-} %>
-
-<%= render "govuk_publishing_components/components/button", text: "Continue", margin_bottom: true %>
+<% content_for :escape_link do %>
+  <%= render "components/escape-link", {
+    text: "Leave this site",
+    href: "https://www.gov.uk/",
```
huwd, line 37: Did we decide that gov.uk home page was the best place to go?
I'd heard a few other sites use the BBC homepage, because it's completely innocuous?
Do we feel the gov.uk homepage offers the same plausible deniability?

app/views/components/docs/escape-link.yml
```diff
@@ -0,0 +1,15 @@
...
+  When JavaScript is disabled the link will point to the passed in reference.
+  When JavaScript is enabled we try to roll them back to the page they visited before starting to use the service.
+  This means no extra requests are needed when trying to hide the current page.
+shared_accessibility_criteria:
+  - link
+examples:
+  default:
+    data:
+      text: Leave this site
+      href: 'https://www.gov.uk/'
```
huwd, line 13: See above

app/views/application/_question_radio_buttons.html.erb
huwd, line 37: Did we decide that gov.uk home page was the best place to go?
I'd heard a few other sites use the BBC homepage, because it's completely innocuous?
Do we feel the gov.uk homepage offers the same plausible deniability?

app/views/application/_question_radio_buttons.html.erb
```diff
@@ -13,22 +13,27 @@
...
-        text: option,
-        id: "option_#{option.parameterize.underscore}",
-        checked: @form_responses[question.to_sym] == option,
-      }
-    end
-} %>
-
-<%= render "govuk_publishing_components/components/button", text: "Continue", margin_bottom: true %>
+<% content_for :escape_link do %>
+  <%= render "components/escape-link", {
```
injms, line 35: Since this is stuck to the bottom of the window, is it worth thinking about where best to put this so that people who use screenreader would be aware of it? Would moving it closer to the start of the markup make it easier to find, whilst not affecting its visible layout?

app/assets/stylesheets/components/_escape-link.scss
```diff
@@ -0,0 +1,25 @@
+.app-c-escape-link {
+  background-color: govuk-colour('white');
+  border-top: 1px solid $govuk-border-colour;
+  bottom: 0;
+  position: sticky;
```
injms, line 5: Fallback for browsers that don't support sticky:

```suggestion
  position: fixed;
  position: sticky;
```

alex-ju: Good point. This requires a 'width: 100%;' on the container. Updated.

app/views/application/_question_radio_buttons.html.erb
alex-ju: Given a screen reader will know if the question is sensible only after reaching the question I consider it's more useful to be able to continue down the page to reach the exit link than going backwards (which usually requires an extra key to be pressed). I wouldn't be too concerned about this as from observed research screen reader users know how to quickly close the current tab/window using shortcuts (sometimes predefined) compared to other users. Open to suggestions if we have more insights on this.

app/views/application/_question_radio_buttons.html.erb
alex-ju: Nope, I worked with gov.uk in the absence of a decision. Happy to change it to anything really.

huwd: Perhaps stick with this for now, and then we can ask for a best practice / product steer and update if needed?

app/views/application/_question_radio_buttons.html.erb
injms: @alex-ju yep - what you say makes sense. Thanks for sharing your insight on this 👍

PR #35, opened 2020-04-08
Set the correct pipeline when updating

richardTowers:


richardTowers:
> Could we not rename the pipeline so that it's consistent with the others?

Then I would have had to set all the secrets again 😭

PR #41, opened 2020-04-09
Route between questions

brucebolt:
This PR carries out a number of tasks:
- Determines which questions a user will see
- Adds redirects to the next question after submitting a valid response
- Adds previous paths to all questions
- Ensures user has answered the urgent medical help question if deep-linked to any part of the form
- Ensures user has answered the filter question (what do you need help with) if deep-linked to any part of the form having already answered the urgent medical help question

Trello card: https://trello.com/c/Cx4lIFW3

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -44,6 +44,12 @@ private
     redirect_to session_expired_path
   end
 
+  def check_first_question_answered
+    unless first_question_seen?
+      redirect_to controller: "urgent_medical_help", action: "show"
+    end
+  end
+
   def check_filter_question_answered
     if questions_to_ask.blank?
       redirect_to controller: "need_help_with", action: "show"
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,8 +7,6 @@ class ApplicationController < ActionController::Base
 
   rescue_from ActionController::InvalidAuthenticityToken, with: :session_expired
 
-  before_action :check_first_question, only: [:show]
-
   def show
     @form_responses = session.to_hash.with_indifferent_access
     respond_to do |format|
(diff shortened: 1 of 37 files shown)


spec/requests/able_to_leave_spec.rb
```diff
@@ -36,10 +40,10 @@
       expect(session[:able_to_leave]).to eq(selected_option)
     end
 
-    xit "redirects to the next question" do
+    it "redirects to the next question" do
       post able_to_leave_path, params: { able_to_leave: selected_option }
 
-      expect(response).to redirect_to(next_question_path)
+      expect(response).to redirect_to(controller: "feel_safe", action: "show")
```
leenagupte, line 58: Minor Nitpick: Can use `redirect_to feel_safe_path`

app/controllers/coronavirus_form/able_to_leave_controller.rb
```diff
@@ -33,6 +33,6 @@ def group
   end
 
   def previous_path
-    "/"
+    previous_question(controller_name)
```
leenagupte, line 38: If this is always the same, it can probably be moved the application controller.

app/controllers/application_controller.rb
```diff
@@ -46,7 +44,9 @@ def session_expired
     redirect_to session_expired_path
   end
 
-  def check_first_question
-    # TODO - do we need to implement this? Or can people be deep linked into the form?
+  def check_filter_question_answered
+    if questions_to_ask.blank?
+      redirect_to controller: "need_help_with", action: "show"
```
leenagupte, line 55: Minor Nitpick: could use `redirect_to need_help_with_path`

leenagupte:
Amazeballs!

PR #43, opened 2020-04-09
Add opt-in usage tracking

injms:
## What

Adds Google Analytics user tracking and a cookie banner to ask people to opt in or out of setting any non-essential cookies.

[Trello card](https://trello.com/c/b1veDIHQ)

This is merrily copied from the excellent work by @bilbof in https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/198

## Why

Adding Google Analytics to this service will help us understand how people are using this form.

Since Google Analytics uses a cookie, a user needs to opt in to before any tracking can be done. The cookie banner provides a consistent way of doing this, as it uses the [cookie component](https://components.publishing.service.gov.uk/component-guide/cookie_banner) from the GOV.UK Publishing Components.

## Visual changes

Cookie banner: 

image

Cookie settings page:

image


app/assets/stylesheets/application.scss
```diff
@@ -2,29 +2,27 @@
 @import 'govuk_publishing_components/govuk_frontend_support';
 @import 'govuk_publishing_components/component_support';
 
-// Components
+// Components@import 'govuk_publishing_components/govuk_frontend_support';
```
huwd, line 5: Should this be on its own line?

injms: Ah yes, it's actually not needed at all... Good spot! Removing it now.

injms: Fixed.

PR #48, opened 2020-04-09
Always show last question

brucebolt:
Regardless of the options selected by the user, we should always show the "are you able to leave home" question.

### DIFF
--- a/app/helpers/questions_helper.rb
+++ b/app/helpers/questions_helper.rb
@@ -1,5 +1,5 @@
 module QuestionsHelper
-  RESULTS_PAGE = "results".freeze
+  FINAL_QUESTION = "able_to_leave".freeze
   FILTER_QUESTION = "need_help_with".freeze
 
   def determine_user_questions(groups)
@@ -18,13 +18,15 @@ module QuestionsHelper
     if current_question_index.nil?
       questions_to_ask.first
     elsif current_question_index == (questions_to_ask.length - 1)
-      RESULTS_PAGE
+      FINAL_QUESTION
     else
       questions_to_ask[current_question_index + 1]
     end
   end
 
   def previous_question(current_question)
+    return questions_to_ask.last if current_question == FINAL_QUESTION
+
     current_question_index = questions_to_ask.index(current_question)
     if current_question_index.zero?
       FILTER_QUESTION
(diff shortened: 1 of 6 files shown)


app/helpers/questions_helper.rb
```diff
@@ -18,15 +18,17 @@ def next_question(current_question)
...
     else
       questions_to_ask[current_question_index + 1]
     end
   end
 
   def previous_question(current_question)
     current_question_index = questions_to_ask.index(current_question)
-    if current_question_index.zero?
+    if current_question_index.nil?
+      questions_to_ask.last
```
leenagupte, line 30: What is this doing?

brucebolt: Allowing us to populate the Back link on the "are you able to leave home" question

leenagupte: So it's basically checking if current_question == FINAL_QUESTION?

brucebolt: No.  If the current question (i.e. `able_to_leave`) is not in `questions_to_ask`, then `.index` will return `nil`, in which case we should use the last question on the back button.

leenagupte: But you could also do it as `return questions_to_ask.last if current_question == FINAL_QUESTION?` before doing all the stuff with `current_question_index`?

brucebolt: That's true.  I'll update this.

app/helpers/questions_helper.rb
```diff
@@ -18,13 +18,15 @@ def next_question(current_question)
...
       questions_to_ask[current_question_index + 1]
     end
   end
 
   def previous_question(current_question)
+    return questions_to_ask.last if current_question == FINAL_QUESTION
+
     current_question_index = questions_to_ask.index(current_question)
     if current_question_index.zero?
       FILTER_QUESTION
```
leenagupte, line 32: So, I guess this could be:
```
return FILTER_QUESTION if current_question == questions_to_ask.first
```
as well?

leenagupte: I feel like I'm understanding this module better now.

brucebolt: Added an additional commit to change the logic there.  Making this explicit does feel like a better thing to do.

leenagupte: Thanks! It looks good ✅

leenagupte: Just seen this: `questions_to_ask[current_question_index + -1]` 🙈  Was that just meant to be `current_question_index - 1`

PR #54, opened 2020-04-09
Store responses

JonathanHallam:
This creates a new table and stores session data in it. 

Trello - https://trello.com/c/f427oVl3/197-store-results-in-postgres

Co-authored-by: Richard Towers <richard.towers@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/controllers/coronavirus_form/able_to_leave_controller.rb
+++ b/app/controllers/coronavirus_form/able_to_leave_controller.rb
@@ -20,11 +20,22 @@ class CoronavirusForm::AbleToLeaveController < ApplicationController
       render controller_path
     else
       update_session_store
+      write_responses
       redirect_to results_path
     end
   end
 
 private
+
+  # We're using this method to reduce the precision of timekeeping so that
+  # responses cannot be correlated with analytics data
+  def time_hour_floor
+    Time.current.beginning_of_hour
+  end
+
+  def write_responses
+    FormResponse.create(form_response: session, created_at: time_hour_floor)
+  end
 
   def update_session_store
     session[:able_to_leave] = @form_responses[:able_to_leave]
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+class ApplicationRecord < ActiveRecord::Base
+  self.abstract_class = true
+end
--- a/app/models/form_response.rb
+++ b/app/models/form_response.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+class FormResponse < ApplicationRecord
+end
--- a/db/migrate/20200409135543_create_form_responses.rb
+++ b/db/migrate/20200409135543_create_form_responses.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class CreateFormResponses < ActiveRecord::Migration[6.0]
+  def change
+    create_table :form_responses do |t|
+      t.jsonb :form_response, default: {}, null: false
+      t.datetime :created_at, null: false
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,9 +10,14 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 0) do
+ActiveRecord::Schema.define(version: 2020_04_09_135543) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+
+  create_table "form_responses", force: :cascade do |t|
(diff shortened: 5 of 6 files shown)


app/controllers/coronavirus_form/able_to_leave_controller.rb
```diff
@@ -20,12 +20,19 @@ def submit
...
+      write_responses
       redirect_to results_path
     end
   end
 
 private
 
+  def write_responses
+    time = Time.current
+    time_hour_floor = time - time.sec - time.min % 60 * 60
```
injms, line 32: Super nice.

brucebolt, line 32: I think this might be one of those rare occasions when a comment in the code could be helpful to explain why this line is important

JonathanHallam: controversial! but agreed, have added this in

JonathanHallam: :doths_cap

issyl0: Definitely.

app/controllers/coronavirus_form/able_to_leave_controller.rb
```diff
@@ -20,12 +20,24 @@ def submit
...
 
+  # We're using this method to reduce the precision of timekeeping so that
+  # responses cannot be correlated with analytics data
+  def time_hour_floor
+    time = Time.current
+    time - time.sec - time.min % 60 * 60
+  end
+
+  def write_responses
+    FormResponse.create({ form_response: session, created_at: time_hour_floor })
```
leenagupte, line 38: I'm surprised linter didn't complain about the curly braces 🤔

app/controllers/coronavirus_form/able_to_leave_controller.rb
```diff
@@ -20,12 +20,24 @@ def submit
       render controller_path
     else
       update_session_store
+      write_responses
```
huwd, line 23: Interesting, I was thinking of moving this to the show method of the results controller.

So instead of writing on the last question, we'd write everytime we load/refresh the results page.

brucebolt: I think the show method wouldn't necessarily be the right place, since we'll be making a GET request.  Maybe the results page should be loaded on a POST request, then it would seem acceptable to write to the database there.

app/controllers/coronavirus_form/able_to_leave_controller.rb
```diff
@@ -20,12 +20,24 @@ def submit
...
     end
   end
 
 private
 
+  # We're using this method to reduce the precision of timekeeping so that
+  # responses cannot be correlated with analytics data
+  def time_hour_floor
+    time = Time.current
+    time - time.sec - time.min % 60 * 60
```
leenagupte, line 34: Have you seen this: https://api.rubyonrails.org/classes/Time.html#method-i-beginning_of_hour ?

leenagupte: Gives you this:

```
[1] pry(main)> time = Time.current
=> Thu, 09 Apr 2020 15:51:03 UTC +00:00
[2] pry(main)> time.beginning_of_hour
=> Thu, 09 Apr 2020 15:00:00 UTC +00:00
[3] pry(main)> 
```

JonathanHallam: well that does exactly what we wanted, nice spot!

JonathanHallam: done!

app/controllers/coronavirus_form/able_to_leave_controller.rb
JonathanHallam: me too, have changed

app/controllers/coronavirus_form/able_to_leave_controller.rb
richardTowers: I think this is the best we're going to be able to come up with tonight.

In most forms there would be a check-your-answers page or something, but this one doesn't have that, so last page it is.

app/controllers/coronavirus_form/able_to_leave_controller.rb
richardTowers: 👨‍🍳💋  - nice! We looked at Ruby's time class, didn't realise Rails had monkey patched it for us!

leenagupte: Rule of thumb with Rails is, if you think there should be a nice way to do something, someone has probably already written it.

PR #56, opened 2020-04-09
Add groups to result links

huwd:
## What

Previously results were nested under questions, this:
- further nests them under groups
- adds titles to the results questions we can use to render the results page.

## Why

Questions are nested within groups, so we've added the groups to the result data along with the relevant titles.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -324,7 +324,7 @@ en:
       title: "Feeling Safe"
       results:
         feel_safe:
-          title: "If you or someone you know are not feeling safe where you live"
+          title: "If you do not feel safe where you live or you’re worried about someone else"
           show_options:
             - "Yes but I’m concerned about the safety of someone else"
             - "No"
@@ -359,7 +359,7 @@ en:
       title: "Paying bills"
       results:
         afford_rent_mortgage_bills:
-          title: "If you're finding it hard to pay rent or bills"
+          title: "If you’re finding it hard to afford rent, your mortgage or bills"
           show_options:
             - "Yes"
             - "Not sure"
@@ -390,7 +390,7 @@ en:
       title: "Getting food"
       results:
         afford_food:
-          title: "If you're finding it hard to afford food"
+          title: "If you’re finding it hard to afford food"
           show_options:
             - "Yes"
             - "Not sure"
@@ -423,7 +423,7 @@ en:
       title: "Being unemployed or not having any work"
       results:
         have_you_been_made_unemployed:
-          title: "If you’ve been made unemployed or put on temporary leave (furloughed)"
+          title: "If you’ve been made unemployed, or put on temporary leave (furloughed)"
           show_options:
             - "I might be soon"
             - "No"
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -242,6 +242,41 @@ en:
               - "I cannot go out because I have a disability"
               - "I’m unable to leave my home for another reason"
   results_link:
+    feel_unsafe:
+      title: "Feeling Safe"
```
leenagupte, line 246: We should be able to get this from the group.

huwd: How do you mean?
Ideally I'd like to keep these as aligned to the keys as possible?

So i can ask the session for the relevant keys then look up the right part of the data structure?

It makes sense to me to pass in one data structure instead of do multiple lookups, if i get this group key out of this part of the YAML i get:
- The question title for the header,
- The key i can use to exclude groups we don't need
- The question headers we'll need for each section
- The result question critiera we can use to exclude questions we don't need
- the results to render

If i did it another way i'd need to splice all that together in the helper no?

huwd: If the downside is it clutters the locale file,

I'm also generally thinking that i really am not worried about how clean this YAML file is (if it's still here in 6th months i'll regret typing this).

But there's that other branch around questions as data, that I think should really clean up and make this all a nicer place for content designers to work with. As there shouldn't be any updates over the weekend, we could prioritize that as first thing after MVP and tidy this up then?

leenagupte: If you have the key name for the group, you can use it to lookup the group title from the locale file, rather than defining it twice and risking them getting out of sync.

leenagupte: This will become clear when the page helper is written.

huwd: ok removed

config/locales/en.yml
```diff
@@ -320,6 +320,41 @@ en:
               - "I’m unable to leave my home for another reason"
             custom_select_error: "Select if you are able to leave your home or not"
   results_link:
+    feel_unsafe:
+      title: "Feeling Safe"
+      results:
+        feel_safe:
+          title: "If you or someone you know are not feeling safe where you live"
```
JonathanHallam: It may not have been you but is this the title of that page?

config/locales/en.yml
```diff
@@ -330,8 +330,8 @@ en:
             - "No"
             - "Not sure"
           links:
-            - text: 'If you’re in immediate danger call <a href="tel:999" class="govuk-link">999</a> and ask for the Police.'
-            - text: 'If you’re in danger and unable to talk on the phone, call <a href="tel:999" class="govuk-link">999</a>, and then press 55.'
+            - text: "If you’re in immediate danger call 999 and ask for the Police."
```
JonathanHallam: Think this is already in the codebase, may need a rebase

config/locales/en.yml
leenagupte: No it's the sub-heading for a set of links for the results in that group. It should be linked to an answer or a group of answers (e.g. Yes and "not sure").

JonathanHallam: oh right, there's some context i missed out on, no worries :)

PR #60, opened 2020-04-09
Update error message

injms:
Update error message for 'Can you get food?' to 'Select yes if you are able to get food'.

### DIFF
--- a/spec/requests/get_food_spec.rb
+++ b/spec/requests/get_food_spec.rb
@@ -62,7 +62,7 @@ RSpec.describe "get-food" do
       post get_food_path
 
       expect(response.body).to have_content(I18n.t("coronavirus_form.groups.getting_food.questions.get_food.title"))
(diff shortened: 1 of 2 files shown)


spec/requests/get_food_spec.rb
```diff
@@ -62,7 +62,7 @@
       post get_food_path
 
       expect(response.body).to have_content(I18n.t("coronavirus_form.groups.getting_food.questions.get_food.title"))
-      expect(response.body).to have_content(I18n.t("coronavirus_form.errors.radio_field", field: I18n.t("coronavirus_form.groups.getting_food.questions.get_food.title").downcase))
+      expect(response.body).to have_content(I18n.t("coronavirus_form.groups.getting_food.questions.get_food.custom_select_error", field: I18n.t("coronavirus_form.groups.getting_food.questions.get_food.title").downcase))
```
brucebolt, line 65: You can also remove the `field` now we're using a custom error
```suggestion
      expect(response.body).to have_content(I18n.t("coronavirus_form.groups.getting_food.questions.get_food.custom_select_error"))
```

injms: Good spot - thanks! I've made the changes and pushed again.

PR #62, opened 2020-04-09
Don't enable basic auth on prod

richardTowers:
Note that this PR won't actually remove the env var, you'll have to run:

```
cf unset-env govuk-coronavirus-find-support REQUIRE_BASIC_AUTH
```

and

```
cf v3-restart govuk-coronavirus-find-support
```

### DIFF
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -114,7 +114,7 @@ jobs:
           CF_SPACE: production
           INSTANCES: 30
           CF_STARTUP_TIMEOUT: 15 # minutes
-          REQUIRE_BASIC_AUTH: "true"
+          REQUIRE_BASIC_AUTH:
           BASIC_AUTH_PASSWORD: ((basic-auth-password))
           AWS_ACCESS_KEY_ID: ((aws-access-key-id-prod))
           AWS_SECRET_ACCESS_KEY: ((aws-secret-access-key-prod))


concourse/pipeline.yml
```diff
@@ -114,7 +114,7 @@ jobs:
           CF_SPACE: production
           INSTANCES: 30
           CF_STARTUP_TIMEOUT: 15 # minutes
-          REQUIRE_BASIC_AUTH: "true"
+          REQUIRE_BASIC_AUTH: "false"
```
issyl0, line 117: Hmm. In the vulnerable people form, `REQUIRE_BASIC_AUTH` is nil (https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/master/concourse/pipeline.yml#L123). This is because in that pipeline, [the deploy to PaaS](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/master/concourse/tasks/deploy-to-govuk-paas.yml#L43-L45) checks if it's set to true or anything, and sets the envvar if so.

That behaviour seems to be the same [in this pipeline](https://github.com/alphagov/govuk-coronavirus-find-support/blob/master/concourse/tasks/deploy-to-govuk-paas.yml#L43-L45), so do we want to do this:

```suggestion
          REQUIRE_BASIC_AUTH:
```

❓

brucebolt: Thanks Issy.  I've updated to remove the value completely.

issyl0: @brucebolt You removed it completely in staging, not production.

PR #66, opened 2020-04-10
Display session expired page on /results to users without session

JonathanHallam:
This will stop users who try to navigate directly to /results from seeing confusing data.

Trello - https://trello.com/c/aQJy0zpB/190-handle-the-results-page-when-the-user-doesnt-have-a-session

### DIFF
--- a/spec/requests/results_spec.rb
+++ b/spec/requests/results_spec.rb
@@ -1,0 +1,8 @@
+RSpec.describe "results" do
+  describe "GET /results" do
+    it "redirects to the session expired path if a user hasn't seen the last question" do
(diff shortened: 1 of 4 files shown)


spec/requests/results_spec.rb
```diff
@@ -0,0 +1,8 @@
+RSpec.describe "afford-rent-mortgage-bills" do
```
brucebolt, line 1:
```suggestion
RSpec.describe "results" do
```

JonathanHallam: :doh, fixed

PR #68, opened 2020-04-10
Fix two content issues

benjamin-mortimer:
**Caution** 
This PR only updates config/locales/en.yml  - I'm not sure how that file interacts with the rest of the app.
I haven't run the app locally (only have standard build Mac) so I don't know if this change will break any tests, for example.

Two small content changes:

L309-11
The wording of the mental health question is slightly confusing - it can read like it's not a 'yes or no' question. Changes the options and adds the word 'either' to make this clearer.

L367
We link to a victim support site for people in Wales - it is in Welsh which some users might not be expecting. Change indicates to the user that the site is in Welsh.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -306,10 +306,10 @@ en:
         title: "Mental health and wellbeing"
         questions:
           mental_health_worries:
-            title: "Are you worried about your mental health or someone else’s mental health?"
+            title: "Are you worried about either your mental health or someone else’s mental health?"
             options:
-              - "Yes"
-              - "No"
+              - "Yes, I am"
+              - "No, I'm not"
               - "Not sure"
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -306,10 +306,10 @@ en:
         title: "Mental health and wellbeing"
         questions:
           mental_health_worries:
-            title: "Are you worried about your mental health or someone else’s mental health?"
+            title: "Are you worried about either your mental health or someone else’s mental health?"
             options:
-              - "Yes"
-              - "No"
+              - "Yes, I am"
+              - "No, I'm not"
```
alex-ju, line 312:
```suggestion
              - "No, I’m not"
```

PR #72, opened 2020-04-10
Correct still_working results title

benjamin-mortimer:
Reads "If you’re still going in to work even though you’re a key worker" - should be, "even though you're _not_ a key worker"


injms:
Merging in as @benjamin-mortimer doesn't have permission to.

benjamin-mortimer:
Cheers!

PR #82, opened 2020-04-10
Add comma

injms:
Add comma to question to make it read better.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -208,7 +208,7 @@ en:
             title: "Do you feel safe where you live?"
             options:
               - "Yes"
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -208,7 +208,7 @@ en:
             title: "Do you feel safe where you live?"
             options:
               - "Yes"
-              - "Yes but I’m concerned about the safety of someone else"
+              - "Yes, but I’m concerned about the safety of someone else"
```
JonathanHallam, line 211: needs adding to the show options too, I'll do that now

PR #89, opened 2020-04-10
Make contractions consistent

hannahjump:
For example you are should become 'you're'

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -362,7 +362,7 @@ en:
             href: "https://womensaid.scot/"
           - text: "Get information for children from Childline"
             href: "https://www.childline.org.uk/"
-          - text: "Get information if you‘re child in Wales"
+          - text: "Get information if you‘re a child in Wales"
             href: "https://www.childcomwales.org.uk/coronavirus/"
           - text: "Contact NSPCC if you‘re concerned about a child"
             href: "https://www.nspcc.org.uk/keeping-children-safe/our-services/nspcc-helpline/"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -418,7 +418,7 @@ en:
             href: "https://www.mygov.scot/school-meals/"
           - text: "Check if you're eligible for the Discretionary Assistance Fund if you live in Wales"
             href: "https://gov.wales/discretionary-assistance-fund-daf"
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -362,11 +362,11 @@ en:
             href: "https://womensaid.scot/"
           - text: "Get information for children from Childline"
             href: "https://www.childline.org.uk/"
-          - text: "Get information if you are child in Wales"
+          - text: "Get information if you‘re child in Wales"
```
huwd, line 365: Should that either be:
- Get information if you‘re a child in Wales, or
- Get information if you‘re child is in Wales

huwd, line 365: Should that either be:
- Get information if you‘re a child in Wales, or
- Get information if your child is in Wales

config/locales/en.yml
```diff
@@ -436,11 +436,11 @@ en:
           - "Yes"
           - "Not sure"
         items:
-          - text: "What to do if you have lost your job"
+          - text: "What to do if you've lost your job"
```
huwd, line 439:
```suggestion
          - text: "What to do if you‘ve lost your job"
```

config/locales/en.yml
```diff
@@ -436,11 +436,11 @@ en:
...
             href: "https://www.gov.uk/guidance/coronavirus-covid-19-what-to-do-if-you-were-employed-and-have-lost-your-job"
           - text: "What to do if you’re laid-off"
             href: "https://www.gov.uk/lay-offs-short-timeworking"
-          - text: "Find out about your rights if you have been dismissed"
+          - text: "Find out about your rights if you've been dismissed"
```
huwd, line 443:
```suggestion
          - text: "Find out about your rights if you‘ve been dismissed"
```

config/locales/en.yml
hannahjump: Good spot 'Get information if you‘re a child in Wales'

config/locales/en.yml
hannahjump: yep, should be the type of apostrophes we use, wasn't intentional!

config/locales/en.yml
hannahjump: yep, should be the type of apostrophes we use, wasn't intentional!

config/locales/en.yml
huwd:
```suggestion
          - text: "Get information if you‘re a child in Wales"
```

PR #95, opened 2020-04-13
Add rake task for extracting statistics

brucebolt:
Today we were asked by the SMT for statistics on how many selections of each option were made for each question.  This involved writing either database queries, or using the Rails console.

Here I have added a basic rake task to output these statistics for a specified day.

Rake task is invoked using (for yesterday's data):
`rake statistics:form_responses`

For a specified day's data:
`rake statistics:form_responses["2020-04-13"]`

The format of the data output is (this information is mocked example data, and not representative of real form submissions):
```
What do you need to find help with because of coronavirus?
 - ["Feeling unsafe"]: 1
 - ["Feeling unsafe", "Paying bills"]: 3
 - ["Getting food"]: 2
 - ["Going in to work"]: 1
 - ["Having somewhere to live"]: 1
 - ["Paying bills"]: 3
 - ["Paying bills", "I’m not sure"]: 1
 - ["Paying bills", "Mental health and wellbeing"]: 1

Do you need urgent medical help?
 - No: 13

Do you feel safe where you live?
 - No: 2
 - Yes: 1
 - Yes, but I’m concerned about the safety of someone else: 1
```

Trello card: https://trello.com/c/ucfW2mvB

### DIFF
--- a/lib/tasks/statistics.rake
+++ b/lib/tasks/statistics.rake
@@ -1,0 +1,44 @@
+namespace :statistics do
+  desc "Get information on all form responses for a specified day, e.g. rake statistics:form_responses[2020-04-13]"
+  task :form_responses, [:date] => [:environment] do |_, args|
+    # Counts responses for each question (as defined in locales) on a given day
+    # and produces output in the following format (fictional data used as an
+    # example):
+    #
+    # What do you need to find help with because of coronavirus?
+    # - ["Feeling unsafe"]: 1
+    # - ["Feeling unsafe", "Paying bills"]: 3
+    # - ["Getting food"]: 2
+    # - ["Going in to work"]: 1
+    # - ["Having somewhere to live"]: 1
+    # - ["Paying bills"]: 3
+    # - ["Paying bills", "I’m not sure"]: 1
+    # - ["Paying bills", "Mental health and wellbeing"]: 1
+    #
+    # Do you need urgent medical help?
+    #  - No: 13
+    #
+    # Do you feel safe where you live?
+    #  - No: 2
+    #  - Yes: 1
+    #  - Yes, but I’m concerned about the safety of someone else: 1
+
+    args.with_defaults(date: Date.yesterday.to_s)
+
+    responses = FormResponse.where(created_at: Date.parse(args.date).all_day)
+
+    questions = I18n.t("coronavirus_form.groups").map { |_, v| v[:questions] }.reduce(:merge)
+
+    questions.each_key do |question|
+      question_text = questions.dig(question, :title)
+      response_count = responses.
+        group("form_response -> '#{question}'").
+        count.
+        reject { |k, _| k.nil? }.
+        sort.
+        map { |k, v| " - #{k}: #{v}" }.
+        join("\n")
+      puts "#{question_text}\n#{response_count}\n\n"
+    end
+  end
+end


richardTowers:
I think we could do a bit more in the database layer here, with something like

```
select
  form_response->'get_food',
  count(*)
from form_responses
group by
  form_response->'get_food';
```

e.g. in staging:

```
+------------+---------+
| ?column?   | count   |
|------------+---------|
| <null>     | 59      |
| "No"       | 150     |
| "Not sure" | 8       |
| "Yes"      | 25      |
+------------+---------+
```

I'm not sure how to active-recordify that, or if there's something elegant we could do to query all the top level keys at once.

brucebolt:
Thanks Richard, that's a much more elegant solution.  I've taken your query and used ActiveRecord to obtain the count values.  I wasn't able to reduce the query to only one, I think we'll still need to query for each question independently.

lib/tasks/statistics.rake
```diff
@@ -0,0 +1,44 @@
...
+    # Do you feel safe where you live?
+    #  - No: 2
+    #  - Yes: 1
+    #  - Yes, but I’m concerned about the safety of someone else: 1
+
+    args.with_defaults(date: Date.yesterday.to_s)
+
+    responses = FormResponse.where(created_at: Date.parse(args.date).all_day)
+
+    questions = I18n.t("coronavirus_form.groups").map { |_, v| v[:questions] }.compact.reduce(:merge)
```
leenagupte, line 30: The vulnerable people form doesn't have any groups. So I think this can be: `I18n.t("coronavirus_form.questions")`

brucebolt: This is for the find support form 😁

leenagupte: 🤦

leenagupte: Ignore me! But the point about tests still stands _she says saving herself_

leenagupte: If you're getting all questions from all groups, which I think you are, then you don't need `.compact.reduce(:merge)` because that gets rid of the `nils`. We used it whne excluding groups without a title.

lib/tasks/statistics.rake
```diff
@@ -0,0 +1,44 @@
...
+    #  - No: 13
+    #
+    # Do you feel safe where you live?
+    #  - No: 2
+    #  - Yes: 1
+    #  - Yes, but I’m concerned about the safety of someone else: 1
+
+    args.with_defaults(date: Date.yesterday.to_s)
+
+    responses = FormResponse.where(created_at: Date.parse(args.date).all_day)
```
leenagupte, line 28: Didn't know about `all_day`. Very nice!

lib/tasks/statistics.rake
brucebolt: It looks like I don't need the `compact`, but I do need the `reduce(:merge)`, otherwise we can get an array containing a single element (a hash of all the questions).

lib/tasks/statistics.rake
```diff
@@ -0,0 +1,44 @@
+namespace :statistics do
+  desc "Get information on all form responses for a specified day, e.g. rake statistics:form_responses[\"2020-04-13\"]"
```
bilbof, line 2:
```suggestion
  desc "Get information on all form responses for a specified day, e.g. rake statistics:form_responses[2020-04-13]"
```

brucebolt: Ooh, I hadn't realised you could drop the quotes

PR #96, opened 2020-04-13
ci/travis: Only run on master

issyl0:
- Now we've got GitHub Actions running for PR builds, we don't need to
  run Travis PR builds at the same time, so we can save a bit of the
  planet by only running the Travis checks on `master`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+branches:
+  only:
+  - master
 language: ruby
 cache: bundler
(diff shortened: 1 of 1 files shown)


issyl0:
> To only build on push events not on pull requests, disable Build on Pull Requests in your repository settings.

[Done](https://travis-ci.org/github/alphagov/govuk-coronavirus-find-support/settings)!

PR #98, opened 2020-04-14
Switch the master build of all forms to use Concourse, not Travis

Rosa-Fox:
- Using Travis meant that we needed to maintain a separate Concourse resource and it's often slow towards the end of the day as that's when the US wakes up and starts using it for all their builds. This leads to queues of builds, and slower deployments for us.

- We can cut out Travis by using Concourse for running tests within the pipeline.

- This commit does not remove Travis, but adds the jobs and resources that we will need. If this is successful then we will remove Travis in a future commit.

Co-authored-by: Huw Diprose <huw.diprose@digital.cabinet-office.gov.uk>

### DIFF
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -76,12 +76,12 @@ jobs:
     serial: true
     build_logs_to_retain: 100
     plan:
-      - get: git-tests-image-deps
+      - get: git-test-image-deps
         trigger: true
       - put: tests-image
         params:
           build: build-tests-image-deps
-          dockerfile: git-tests-image-deps/features/Dockerfile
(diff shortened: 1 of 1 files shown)


huwd:
Plan is:
- Check these resources and Jobs function properly
- Remove Travis and knit these into the pipeline
- Optimise the docker images to ensure we're caching as much as possible and things run quick.

huwd:
@richardTowers no not yet, We're being cautious about deploying these, want to make sure they work OK alongside the existing stuff. Jobs and resources all work correctly.

Then we'll do a follow up to remove travis and knit this into the triggers.

Then a final one to do an optimization pass, probably add a new docker image with build steps so things cache well?

PR #102, opened 2020-04-15
MVP export of aggregate form response data

brucebolt:
This in a MVP attempt to allow access to aggregated form response data without needing a developer to run a rake task.  A range of dates can be exported, with the data broken down by day.

The data can be accessed by making a GET request with a relevant header, e.g. using curl:
```curl -XGET -H "Accept: text/csv" "<path-to-service>/data-export?start_date=2020-04-08&end_date=2020-04-15"```

Whilst this exports CSV, it can easily be changed to JSON by changing the response type and not calling the `produce_csv` method, should this be required to be made in a live API at some future time.

The ActiveRecord query looks complex, but we are utilising Postgres's JSONB query and counts to avoid the need for the Rails application to retrieve all the data then do the sorting (this is all handled by Postgres and we just receive the count values).  This reduces the amount of data transfer and compute required by the app.

Example of how the data appears (all data is mocked up, and not representative of actual user responses):
```
What do you need to find help with because of coronavirus?,I’m not sure,2020-04-09,1
What do you need to find help with because of coronavirus?,Feeling unsafe,2020-04-10,1
What do you need to find help with because of coronavirus?,Getting food,2020-04-10,2
What do you need to find help with because of coronavirus?,Going in to work,2020-04-10,1
What do you need to find help with because of coronavirus?,Having somewhere to live,2020-04-10,1
Do you need urgent medical help?,No,2020-04-09,1
Do you need urgent medical help?,No,2020-04-10,13
Do you need urgent medical help?,No,2020-04-15,1
Do you feel safe where you live?,No,2020-04-10,3
Do you feel safe where you live?,Yes,2020-04-10,5
Do you feel safe where you live?,"Yes, but I’m concerned about the safety of someone else",2020-04-10,1
Do you feel safe where you live?,Yes,2020-04-15,1
"Are you finding it hard to afford rent, your mortgage or bills?",No,2020-04-10,4
"Are you finding it hard to afford rent, your mortgage or bills?",Yes,2020-04-10,4
"Are you finding it hard to afford rent, your mortgage or bills?",Yes,2020-04-15,1
Are you finding it hard to afford food?,No,2020-04-10,5
Are you finding it hard to afford food?,Yes,2020-04-10,2
Are you finding it hard to afford food?,No,2020-04-15,1
```

Trello card: https://trello.com/c/8U3PfrO6

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,3 +57,13 @@ which is configured in [concourse/pipeline.yml](concourse/pipeline.yml).
 
 The concourse pipeline has credentials for the `govuk-forms-deployer` user in
 GOV.UK PaaS. This user has the SpaceDeveloper role, so it can `cf push` the application.
+
+## Exporting form response data
+
+Aggregated anonymous daily data can be extracted from the application for any
+specified time period using the `/data-export` route.  Output is available in
+CSV format.
+
+For example, to get data for a five day period starting 10th April 2020:
+
+     <path-to-service>/data-export.csv?start_date=2020-04-08&end_date=2020-04-15
--- a/app/controllers/coronavirus_form/data_export_controller.rb
+++ b/app/controllers/coronavirus_form/data_export_controller.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+require "csv"
+
+class CoronavirusForm::DataExportController < ApplicationController
+  def show
+    respond_to do |format|
+      format.html { render controller_path }
+      format.csv { render csv: produce_csv(usage_statistics(params[:start_date], params[:end_date])), filename: "data-export" }
+    end
+  end
+
+  def usage_statistics(start_date, end_date)
+    questions = I18n.t("coronavirus_form.groups").map { |_, v| v[:questions] }.reduce(:merge)
+
+    start_date = Date.parse(sanitize(start_date)).beginning_of_day
+    end_date = Date.parse(sanitize(end_date)).end_of_day
+
+    results = {}
+    questions.each_key do |question|
+      question_text = questions.dig(question, :title)
+      counts = FormResponse
+        .where(created_at: start_date..end_date)
+        .select(Arel.sql(%{created_at::date, form_response -> '#{question}', COUNT(*)}))
+        .group(Arel.sql(%{created_at::date, form_response -> '#{question}'}))
+        .pluck(Arel.sql(%{created_at::date, form_response -> '#{question}', COUNT(*)}))
+        .reject { |form_responses| form_responses[1].nil? }
+        .map do |form_responses|
+          {
+            date: form_responses[0].strftime("%Y-%m-%d"),
+            response: Array(form_responses[1]).join(";"),
+            count: form_responses[2],
+          }
+        end
+      results[question_text] = counts if counts.present?
+    end
+    results
+  end
+
+private
+
+  def produce_csv(results)
+    csv_data = CSV.generate do |csv|
+      csv << %w(question answer date count)
+      results.each do |question, value|
+        value.each do |k|
+          csv << [
+            question,
+            k[:response],
+            k[:date],
+            k[:count],
+          ]
+        end
+      end
+    end
+
+    csv_data
+  end
+end
--- a/spec/controllers/coronavirus_form/data_export_controller_spec.rb
+++ b/spec/controllers/coronavirus_form/data_export_controller_spec.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+RSpec.describe CoronavirusForm::DataExportController, type: :controller do
+  subject(:instance) { described_class.new }
+
+  describe "#usage_statistics" do
+    let(:start_date) { "2020-04-10" }
+    let(:end_date) { "2020-04-15" }
+
+    before do
+      FormResponse.create(form_response: { able_to_leave: "Yes", get_food: "Yes" }, created_at: "2020-04-10 10:00:00")
+      FormResponse.create(form_response: { able_to_leave: "Yes", get_food: "No" }, created_at: "2020-04-10 10:00:00")
+      FormResponse.create(form_response: { able_to_leave: "Yes", get_food: "Yes" }, created_at: "2020-04-11 10:00:00")
+      FormResponse.create(form_response: { able_to_leave: "No", get_food: "No" }, created_at: "2020-04-11 10:00:00")
+    end
+
+    it "returns a hash containing aggregated form responses" do
+      expected_response = {
+        "Are you able to leave your home if absolutely necessary?" => [
+          {
+            response: "Yes",
+            date: "2020-04-10",
+            count: 2,
+          },
+          {
+            response: "No",
+            date: "2020-04-11",
+            count: 1,
+          },
+          {
+            response: "Yes",
+            date: "2020-04-11",
+            count: 1,
+          },
+        ],
+        "Are you able to get food?" => [
+          {
+            response: "No",
+            date: "2020-04-10",
+            count: 1,
+          },
+          {
+            response: "Yes",
+            date: "2020-04-10",
+            count: 1,
+          },
+          {
+            response: "No",
+            date: "2020-04-11",
+            count: 1,
+          },
+          {
+            response: "Yes",
+            date: "2020-04-11",
+            count: 1,
+          },
+        ],
+      }
+
+      expect(instance.usage_statistics(start_date, end_date)).to eq(expected_response)
+    end
+  end
+end
--- a/config/initializers/csv_renderer.rb
+++ b/config/initializers/csv_renderer.rb
@@ -1,0 +1,5 @@
+ActionController::Renderers.add :csv do |obj, options|
+  filename = options.fetch(:filename, "data")
+  content = obj.respond_to?(:to_csv) ? obj.to_csv : obj.to_s
+  send_data content, type: Mime[:csv], disposition: "attachment; filename=#{filename}.csv"
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -80,6 +80,8 @@ Rails.application.routes.draw do
     get "/clear-session", to: "session#delete"
 
     get "/session-expired", to: "session_expired#show"
+
+    get "/data-export", to: "data_export#show"
   end
 
   mount GovukPublishingComponents::Engine, at: "/component-guide"
--- a/spec/requests/data_export_spec.rb
+++ b/spec/requests/data_export_spec.rb
@@ -1,0 +1,34 @@
+RSpec.describe "data-export" do
+  let(:start_date) { "2020-04-10" }
+  let(:end_date) { "2020-04-15" }
+
+  before do
+    FormResponse.create(form_response: { able_to_leave: "Yes", get_food: "Yes" }, created_at: "2020-04-10 10:00:00")
+    FormResponse.create(form_response: { able_to_leave: "Yes", get_food: "No" }, created_at: "2020-04-10 10:00:00")
+    FormResponse.create(form_response: { able_to_leave: "Yes", get_food: "Yes" }, created_at: "2020-04-11 10:00:00")
+    FormResponse.create(form_response: { able_to_leave: "No", get_food: "No" }, created_at: "2020-04-11 10:00:00")
+  end
+
+  describe "GET /able-to-leave" do
+    let(:expected_lines) do
+      [
+        %{question,answer,date,count},
+        %{Are you able to leave your home if absolutely necessary?,Yes,2020-04-10,2},
+        %{Are you able to leave your home if absolutely necessary?,Yes,2020-04-11,1},
+        %{Are you able to leave your home if absolutely necessary?,No,2020-04-11,1},
+        %{Are you able to get food?,Yes,2020-04-10,1},
+        %{Are you able to get food?,No,2020-04-10,1},
+        %{Are you able to get food?,Yes,2020-04-11,1},
+        %{Are you able to get food?,No,2020-04-11,1},
+      ]
+    end
+
+    it "shows all expected responses in CSV format" do
+      get data_export_path, params: { start_date: start_date, end_date: end_date }, headers: { "HTTP_ACCEPT" => "text/csv" }
+
+      expected_lines.each do |line|
+        expect(response.body).to have_content(line)
+      end
+    end
+  end
+end

### CODE REVIEW

File: app/controllers/coronavirus_form/data_export_controller.rb [hunk 1]
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::DataExportController < ApplicationController
+  def show
+    render plain: produce_csv(usage_statistics)
```
leenagupte, line 5: Could a `respond_to` be used here to render the csv? See: https://apidock.com/rails/ActionController/MimeResponds/InstanceMethods/respond_to and https://api.rubyonrails.org/classes/ActionController/Renderers.html

app/controllers/coronavirus_form/data_export_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    results = {}
+    questions.each_key do |question|
+      question_text = questions.dig(question, :title)
+      counts = FormResponse.
+        where(created_at: start_date..end_date).
+        select(%{created_at::date, form_response -> '#{question}', COUNT(*)}).
+        group(%{created_at::date, form_response -> '#{question}'}).
+        pluck(%{created_at::date, form_response -> '#{question}', COUNT(*)}).
+        reject { |v| v[1].nil? }.
```
leenagupte, line 24: If v is the group, what's v[1]?

app/controllers/coronavirus_form/data_export_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+        map do |v|
```
leenagupte, line 25: Is this the same `v` as above?

app/controllers/coronavirus_form/data_export_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+  def usage_statistics
+    questions = I18n.t("coronavirus_form.groups").map { |_, v| v[:questions] }.reduce(:merge)
+
+    start_date = Date.parse(sanitize(params[:start_date])).beginning_of_day
+    end_date = Date.parse(sanitize(params[:end_date])).end_of_day
+
+    results = {}
+    questions.each_key do |question|
+      question_text = questions.dig(question, :title)
+      counts = FormResponse.
```
leenagupte, line 19: This is a real nitpick, and I'm surprised the linter didn't catch it, but periods should be at the beginning of of the next line, e.g. `.where......`

app/controllers/coronavirus_form/data_export_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+            response: Array(v[1]).join(";"),
+            count: v[2],
+          }
+        end
+      results[question_text] = counts
+    end
+    results
+  end
+
+  def produce_csv(results)
```
leenagupte, line 37: Ruby has a CSV class that could be worth investigating: https://ruby-doc.org/stdlib-2.6.1/libdoc/csv/rdoc/CSV.html

app/controllers/coronavirus_form/data_export_controller.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::DataExportController < ApplicationController
+  def show
+    render plain: produce_csv(usage_statistics)
+  end
+
+private
+
+  def usage_statistics
```
leenagupte, line 10: I would be great if this could have it's own test, so if we need expand the formats to include other types, like json, we'd have confidence in the results being extracted.

app/controllers/coronavirus_form/data_export_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    end_date = Date.parse(sanitize(params[:end_date])).end_of_day
+
+    results = {}
+    questions.each_key do |question|
+      question_text = questions.dig(question, :title)
+      counts = FormResponse.
+        where(created_at: start_date..end_date).
+        select(%{created_at::date, form_response -> '#{question}', COUNT(*)}).
+        group(%{created_at::date, form_response -> '#{question}'}).
+        pluck(%{created_at::date, form_response -> '#{question}', COUNT(*)}).
```
leenagupte, line 23: These statements are throwing a fair few deprecation warnings:
```
DEPRECATION WARNING: Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s): "created_at::date, form_response -> 'able_to_leave', COUNT(*)". Non-attribute arguments will be disallowed in Rails 6.1. This method should not be called with user-provided values, such as request parameters or model attributes. Known-safe values can be passed by wrapping them in Arel.sql(). (called from block in usage_statistics at /Users/leenagupte/govuk/govuk-coronavirus-find-support/app/controllers/coronavirus_form/data_export_controller.rb:23)
```
There are a handful of examples in alphagov of using `Arel.sql` that we can "borrow" from: https://github.com/search?p=1&q=org%3Aalphagov+Arel.sql&type=Code

app/controllers/coronavirus_form/data_export_controller.rb
brucebolt: It's the second column returned by the select statement, i.e. the response to the question.  I'm removing `nil` since the response includes `nil` where a user never answered a question.

app/controllers/coronavirus_form/data_export_controller.rb
brucebolt: Yes, but with the `nil` responses now removed

app/controllers/coronavirus_form/data_export_controller.rb
brucebolt: Good find.  I'll use `Arel.sql` where required.

app/controllers/coronavirus_form/data_export_controller.rb
brucebolt: I think this is a much nicer approach.  Will update the code to have a CSV renderer and `respond_to` the relevant MIME type.

app/controllers/coronavirus_form/data_export_controller.rb
brucebolt: Thanks, I'll use this.

app/controllers/coronavirus_form/data_export_controller.rb
brucebolt: Will add a unit test for this method.

brucebolt:
Thank you for the review @leenagupte.  I have addressed all your comments and updated the README to give details on how to query this endpoint.

README.md
```diff
@@ -57,3 +57,13 @@ which is configured in [concourse/pipeline.yml](concourse/pipeline.yml).
...
+
+## Exporting form response data
+
+Aggregated anonymous daily data can be extracted from the application for any
+specified time period using the `/data-export` route.  Output is available in
+CSV format.
+
+For example, to get data for a five day period starting 10th April 2020:
+
+     curl -XGET -H "Accept: text/csv" "<path-to-service>/data-export?start_date=2020-04-08&end_date=2020-04-15"
```
richardTowers, line 69: There's no need for the `-XGET` - GET is the default.

```suggestion
     curl -H "Accept: text/csv" "<path-to-service>/data-export?start_date=2020-04-08&end_date=2020-04-15"
```

spec/controllers/coronavirus_form/data_export_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    let(:end_date) { "2020-04-15" }
+
+    before do
+      FormResponse.create(form_response: { able_to_leave: "Yes", get_food: "Yes" }, created_at: "2020-04-10 10:00:00")
+      FormResponse.create(form_response: { able_to_leave: "Yes", get_food: "No" }, created_at: "2020-04-10 10:00:00")
+      FormResponse.create(form_response: { able_to_leave: "Yes", get_food: "Yes" }, created_at: "2020-04-11 10:00:00")
+      FormResponse.create(form_response: { able_to_leave: "No", get_food: "No" }, created_at: "2020-04-11 10:00:00")
+    end
+
+    it "returns a JSON representation of the form responses" do
```
richardTowers, line 19: Technically it returns a ruby hash, rather than a JSON representation.

app/controllers/coronavirus_form/data_export_controller.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require "csv"
+
+class CoronavirusForm::DataExportController < ApplicationController
+  def show
+    respond_to do |format|
+      format.html { render controller_path }
```
richardTowers, line 8: Will this just be an empty page? Should we think about maybe having some kind of view?

README.md
leenagupte, line 69: Thanks for adding this 🎉. The url needs `.csv` added to it to get the CSV download. 
i.e. `/data-export.csv?start_date=2020-04-08&end_date=2020-04-15"`

app/controllers/coronavirus_form/data_export_controller.rb
brucebolt: Yes, it will be a blank page.  Should we have some sort of content here?

README.md
brucebolt: I didn't know Rails let you do this.  That means I can remove the accept header too.

spec/controllers/coronavirus_form/data_export_controller_spec.rb
brucebolt: Ah, yes.  Updated the test description.

app/controllers/coronavirus_form/data_export_controller.rb
brucebolt: I could just remove the HTML line here.  Although we'd get raise an exception if users try navigating in their browser without the `.csv` suffix or correct accept header.

leenagupte: An interesting thing happens here. If you accidentally go to `/data-export` without the `.csv` an empty page is rendered:

Screenshot 2020-04-16 at 11 07 17

I don't see this as a blocker, but it is interesting.

app/controllers/coronavirus_form/data_export_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      format.csv { render csv: produce_csv(usage_statistics(params[:start_date], params[:end_date])), filename: "data-export" }
```
leenagupte, line 9: Nitpick, perhaps use a do/end block? e.g.
```
format.csv do
  render csv: produce_csv(usage_statistics(params[:start_date], params[:end_date])),
                      filename: "data-export"
end
```

app/controllers/coronavirus_form/data_export_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    start_date = Date.parse(sanitize(start_date)).beginning_of_day
+    end_date = Date.parse(sanitize(end_date)).end_of_day
+
+    results = {}
+    questions.each_key do |question|
+      question_text = questions.dig(question, :title)
+      counts = FormResponse
+        .where(created_at: start_date..end_date)
+        .select(Arel.sql(%{created_at::date, form_response -> '#{question}', COUNT(*)}))
```
leenagupte, line 24: What do the `%` do?

app/controllers/coronavirus_form/data_export_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+        .group(Arel.sql(%{created_at::date, form_response -> '#{question}'}))
+        .pluck(Arel.sql(%{created_at::date, form_response -> '#{question}', COUNT(*)}))
+        .reject { |form_responses| form_responses[1].nil? }
```
leenagupte, line 27: Thank you! `form_responses` is a good name.
Nitpick, I think `form_responses[1].nil?` could be written as `form_responses.second.nil?`.

app/controllers/coronavirus_form/data_export_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+        .map do |form_responses|
+          {
+            date: form_responses[0].strftime("%Y-%m-%d"),
+            response: Array(form_responses[1]).join(";"),
+            count: form_responses[2],
```
leenagupte, line 32: Not a blocker, but here too, you can use `first`, `second` and `third`.

spec/controllers/coronavirus_form/data_export_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    let(:end_date) { "2020-04-15" }
+
+    before do
+      FormResponse.create(form_response: { able_to_leave: "Yes", get_food: "Yes" }, created_at: "2020-04-10 10:00:00")
+      FormResponse.create(form_response: { able_to_leave: "Yes", get_food: "No" }, created_at: "2020-04-10 10:00:00")
+      FormResponse.create(form_response: { able_to_leave: "Yes", get_food: "Yes" }, created_at: "2020-04-11 10:00:00")
+      FormResponse.create(form_response: { able_to_leave: "No", get_food: "No" }, created_at: "2020-04-11 10:00:00")
+    end
+
+    it "returns a hash containing aggregated form responses" do
```
leenagupte, line 19: Thanks for adding this.

app/controllers/coronavirus_form/data_export_controller.rb
richardTowers: I think an exception is better than a blank page. Or we could render an HTML page with a link to the `.csv` route?

brucebolt: Turns out Rails will return `406 Not Acceptable` (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406) if we don't specify anything to render here.  That feels like the correct response code, so I'll use the default behaviour.

app/controllers/coronavirus_form/data_export_controller.rb
brucebolt: This looks better.  Will update.

app/controllers/coronavirus_form/data_export_controller.rb
brucebolt: It creates a string (https://docs.ruby-lang.org/en/2.0.0/syntax/literals_rdoc.html#label-Percent+Strings).  Much like double quotes, but lets you include quotes in the string without needing to escape them.  I used this since an earlier iteration of the query included quote marks, but these can now be removed.

app/controllers/coronavirus_form/data_export_controller.rb
brucebolt: Sure, will change this

brucebolt:
It seems the latest changes I pushed to this branch before lunch never got merged when I clicked the merge button 😕.  Have opened another PR to make those changes: https://github.com/alphagov/govuk-coronavirus-find-support/pull/109.

PR #105, opened 2020-04-15
Adds test image and "quality checks" to concourse

huwd:
## What
image

- Sets up a new test docker image, rebuilt  whenever we change the `.ruby_file`, `Gemfile` or `Gemfile.lock`
- uses that image to:
  - start the postgres service running as a daemon
  - create the appropriate roles
  - setup and migrate the database
  - run bundle exec rake

## Why
We want a way to run tests within concourse so we can do continuious delivery that ships conditional on passing tests.

Whilst we have that set up currently with Travis, there is a potential race condition where the deployed master and tested branch could diverge.

Adding tests to the pipeline helps prevent deploying code which does not meet quality checks.

## Next steps
Now we know this passes, we can knit this into the pipeline itself,
- removing travis as the trigger and replacing it with any update to the master branch
- running quality checks first in the pipeline and making all subsequent steps conditional on passes

### DIFF
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -105,9 +105,12 @@ jobs:
             args:
               - -c
               - |
-                set -ue
-                cat .git/committer > ../committer-details/last-committer
-                postgres -p 5432 -D /usr/local/var/postgres >&1 &
+                set -eu
+                service postgresql start
+                su - postgres -c "psql -c \"create role root with createdb login password 'password';\""
+                export DATABASE_URL="postgres://root:password@localhost:5432/coronavirus_form_development"
+                bundle exec rails db:setup
+                bundle exec rails db:migrate
                 bundle exec rake
 
   - name: build-feature-tests-image
--- a/features/Dockerfile
+++ b/features/Dockerfile
@@ -1,27 +1,32 @@
 #========================================================================================
-# This Dockerfile creates a container image which contains ruby and google chrome,
-# as well as the application source and bundled dependencies.
+# This Dockerfile creates a container image which contains Ruby,
+# Google Chrome and Postgres, as well as the application source and bundled dependencies.
 #
-# This is needed for running the feature tests in a container environment like concourse.
+# This is needed for running the feature tests in a container environment like Concourse.
 #========================================================================================
 FROM ruby:2.6.5-buster
 
+# Updates all debian dependencies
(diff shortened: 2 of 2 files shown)


concourse/pipeline.yml
```diff
@@ -105,10 +105,12 @@ jobs:
...
               - -c
               - |
-                set -ue
-                cat .git/committer > ../committer-details/last-committer
-                postgres -p 5432 -D /usr/local/var/postgres >&1 &
-                bundle exec rake
+                set -eu
+                service postgresql start
+                su - postgres -c "psql -c \"create role root with createdb login password 'password';\""
+                DATABASE_URL="postgres://root:password@localhost:5432/coronavirus_form_development" bundle exec rails db:setup
```
issyl0, line 111: We can set this once in `params` before the `run`:

```yaml
params:
        DATABASE_URL: "postgres://root:password@localhost:5432/coronavirus_form_development
```

concourse/pipeline.yml
```diff
@@ -105,10 +105,12 @@ jobs:
...
+                DATABASE_URL="postgres://root:password@localhost:5432/coronavirus_form_development" bundle exec rails db:migrate
+                DATABASE_URL="postgres://root:password@localhost:5432/coronavirus_form_development" bundle exec rake
```
Rosa-Fox, line 113: Looks good! Only comment/question was if it would be possible to do `export DATABASE_URL="postgres://root:password@localhost:5432/coronavirus_form_development"` so as not to repeat it? or does it need to be defined for the rails commands?

huwd: yeah much better idea, will give it a go!

concourse/pipeline.yml
huwd: What does that do? Does that set it as a env variable for everything in run?
Or does it prefix it as a parameter to all those calls?

issyl0: The former.

huwd: Does job have a `params:` ? just been reading up [in the docs](https://concourse-ci.org/jobs.html#schema.job) I can see params in the steps schema, but not in jobs :thinking:

issyl0: Oh! Sorry. 😩

huwd: It might well do, not actually tested it, and i find these docs well confusing.
Was wondering if you'd seen it used somewhere before?

PR #107, opened 2020-04-16
Remove cucumber tests

JonathanHallam:
The cucumber tests have been replaced with rspec and we no longer have the cucumber gem. This removes the tests themselves as they are no longer run.

Move to Rspec - https://github.com/alphagov/govuk-coronavirus-find-support/pull/94
Removal of cucumber gem - https://github.com/alphagov/govuk-coronavirus-find-support/pull/106


issyl0:
The Dockerfile is in use for the `master` build steps from #105 (cc @huwd) - I guess we could move it to `concourse/`?

Rosa-Fox:
> The Dockerfile is in use for the `master` build steps from #105 (cc @huwd) - I guess we could move it to `concourse/`?

As well as moving the Docker file, it would be good to remove the `govuk-coronavirus-find-support-feature-tests` from the pipeline as `run-quality-checks` will run the rspec feature specs via the `rake` command. I am conscious over overlapping with @huwd so should I make a separate card for this? Should be quick to do once the other pipeline stuff is merged?

huwd:
No prob, moved the dockerfile and should have tidied up the other feature test stuff here: https://github.com/alphagov/govuk-coronavirus-find-support/pull/108

Rosa-Fox:
Amazing, thanks @huwd! Have rebased your changes from master so this is ready for re review :)

PR #110, opened 2020-04-16
Clear session when using emergency exit button

huwd:
Trello: https://trello.com/c/4JRtGN5p/240-clear-the-users-session-when-they-click-leave-this-site

**draft pending huw doing a manual test to be super sure**

## What
Instead of having the emergency button do directly to a target, route it through the clear session route, which can then redirect them onwards based on a param

## Why
There are fears that if a user clicks the emergency exit button, their session will remain. So if anyone revisists the site, their answers will still be present.

We've decided this presents a risk to some users and a click to the link should clear their session when they redirect.

### DIFF
--- a/app/controllers/coronavirus_form/session_controller.rb
+++ b/app/controllers/coronavirus_form/session_controller.rb
@@ -3,6 +3,10 @@
 class CoronavirusForm::SessionController < ApplicationController
   def delete
     reset_session
-    redirect_to "/"
+    if params[:ext_r] == "true"
+      redirect_to I18n.t("leave_this_website.link_redirect_to")
+    else
+      redirect_to "/"
+    end
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -108,7 +108,8 @@ en:
       <p class="govuk-body">We do this for your security. We've deleted all the details you entered to protect your data.</p>
   leave_this_website:
     link_text: "Leave this site"
-    link_href: "https://www.bbc.co.uk/"
+    link_href: "clear-session?ext_r=true"
+    link_redirect_to: "https://www.bbc.co.uk/"
   get_help_from_nhs:
     title: Get urgent help from the NHS now
     link_text: Go to NHS 111 online
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -108,7 +108,7 @@ en:
       <p class="govuk-body">We do this for your security. We've deleted all the details you entered to protect your data.</p>
   leave_this_website:
     link_text: "Leave this site"
-    link_href: "clear-session?ext_r=true"
+    link_href: "/clear-session?ext_r=true"
     link_redirect_to: "https://www.bbc.co.uk/"
   get_help_from_nhs:
     title: Get urgent help from the NHS now
--- a/spec/requests/clear_session_spec.rb
+++ b/spec/requests/clear_session_spec.rb
@@ -2,24 +2,47 @@ RSpec.describe "clear-session" do
   let(:selected_no) { I18n.t("coronavirus_form.groups.help.questions.urgent_medical_help.options").last }
 
   describe "GET /clear-session" do
-    context "with session data" do
+    context "without a redirect parameter" do
+      context "with session data" do
+        it "clears the session id" do
+          post urgent_medical_help_path, params: { urgent_medical_help: selected_no }
+          initial_session_id = session["session_id"]
+          get clear_session_path
+          expect(session["session_id"]).to_not eq(initial_session_id)
+        end
+
+        it "clears responses held in the session" do
+          post urgent_medical_help_path, params: { urgent_medical_help: selected_no }
+          expect(session[:urgent_medical_help]).to eq(selected_no)
+          get clear_session_path
+          expect(session[:have_you_been_made_unemployed]).to be_nil
+        end
+
+        it "redirects the user to root" do
+          get clear_session_path
+          expect(response).to redirect_to("/")
+        end
+      end
+    end
+
+    context "with an external redirect parameter" do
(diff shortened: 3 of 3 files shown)


app/controllers/coronavirus_form/session_controller.rb
```diff
@@ -3,6 +3,6 @@
 class CoronavirusForm::SessionController < ApplicationController
   def delete
     reset_session
-    redirect_to "/"
+    redirect_to params[:r] || "/"
```
richardTowers, line 6: We'll need to make sure we don't have an [Open Redirect](https://cwe.mitre.org/data/definitions/601.html) here.

huwd: ah, hummm right.

So i could post on twitter https://gov.uk/clear-session?r=https://huws-evil-site-of-doom.com
and i'd get all the nice meta data stuff from our site but fling them onto my evil phising site?

huwd: humm ok not good, will perhaps just make it a switch and hard code in a single redirect location

huwd: like internal = gov.uk root, external = bbc.co.uk or something?

huwd: @richardTowers had a second pass, let me know if that might work instead?

app/controllers/coronavirus_form/session_controller.rb
```diff
@@ -3,6 +3,7 @@
 class CoronavirusForm::SessionController < ApplicationController
   def delete
     reset_session
-    redirect_to "/"
+    redirect_to I18n.t("leave_this_website.link_redirect_to") if params[:ext_r] == true
+    redirect_to "/" if params[:ext_r] != true
```
richardTowers, line 7: Yeah, this looks good from a security perspective.

I'd probably suggest and if / else here, instead of the inline ifs.

Also I wonder if `/clear-session?leave-this-site=true` would be nicer than `/clear-session?ext_r=true`

app/controllers/coronavirus_form/session_controller.rb
richardTowers: Yep, your understanding of the issue is correct, and your new implementation looks good to me.

These issues sometimes pop up with login flows, where it's like `/login?on_success_redirect=some-phishing-site.com`, which is particularly nasty. It would be a bit less bad than that in this case, but still worth avoiding.

app/controllers/coronavirus_form/session_controller.rb
huwd: yeah i wondered about that, was thinking as this shows up in a users history, perhaps this is the one time where a less nice obscure param is ok?
I figured clear session is a bit of a nerd-term anyway?

huwd: also turns out params are strings not booleans...

app/controllers/coronavirus_form/session_controller.rb
```diff
@@ -3,6 +3,7 @@
 class CoronavirusForm::SessionController < ApplicationController
   def delete
     reset_session
-    redirect_to "/"
+    redirect_to I18n.t("leave_this_website.link_redirect_to") if params[:ext_r] == "true"
+    redirect_to "/" if params[:ext_r] != "true"
```
richardTowers, line 7:
```suggestion
    if params[:ext_r] == "true"
        redirect_to I18n.t("leave_this_website.link_redirect_to")
    else
        redirect_to "/"
    end
```

huwd: yeah cool, makes more sense. amended

app/controllers/coronavirus_form/session_controller.rb
huwd: Right, good catch. One for my TIL list!

app/controllers/coronavirus_form/session_controller.rb
leenagupte: Pretty sure you didn't need the second if. I think this could have been written as:

```
redirect_to I18n.t("leave_this_website.link_redirect_to") if params[:ext_r] == "true"
redirect_to "/"
```

spec/requests/clear_session_spec.rb
```diff
@@ -2,24 +2,47 @@
...
 
       it "clears responses held in the session" do
         post urgent_medical_help_path, params: { urgent_medical_help: selected_no }
         expect(session[:urgent_medical_help]).to eq(selected_no)
-        get clear_session_path
+        get clear_session_path, params: { ext_r: true }
         expect(session[:have_you_been_made_unemployed]).to be_nil
       end
 
       it "redirects the user to root" do
```
richardTowers, line 43:
```suggestion
      it "redirects the user to an external website" do
```

config/locales/en.yml
```diff
@@ -108,7 +108,8 @@ en:
       <p class="govuk-body">We do this for your security. We've deleted all the details you entered to protect your data.</p>
   leave_this_website:
     link_text: "Leave this site"
-    link_href: "https://www.bbc.co.uk/"
+    link_href: "clear-session?ext_r=true"
```
bilbof, line 111: Does this need to be "/clear-session?ext_r=true"? Also, I wonder if using `clear_session_path(ext_r: true)` works here?

richardTowers: Yeah, I think you're right that it should have a leading `/`.

I don't think the `clear_session_path` helper will work here because we're not in a ruby context, but we could do that in the view instead of reading the href from the localisation file.

richardTowers, line 111:
```suggestion
    link_href: "/clear-session?ext_r=true"
```

PR #112, opened 2020-04-16
Fix concourse should use renamed git-master resource in files

huwd:
## What
I renamed a resource and missed a couple of references


huwd:
https://github.com/alphagov/govuk-coronavirus-find-support/pull/111

PR #122, opened 2020-04-17
Are you able to leave home

hannahjump:
https://trello.com/c/SV0Kyh1O/262-iterate-content-on-are-you-able-to-leave-your-home

Changed: 
- questions
- options

### DIFF
--- a/spec/support/fill_in_the_form_steps.rb
+++ b/spec/support/fill_in_the_form_steps.rb
@@ -127,7 +127,7 @@ module FillInTheFormSteps
   def and_is_not_able_to_leave_home_if_absolutely_necessary
     expect(page).to have_content(I18n.t("coronavirus_form.groups.leave_home.questions.able_to_leave.title"))
 
-    choose "I cannot go out because I have coronavirus, or someone in my household does"
+    choose "I should not leave home because I have coronavirus symptoms, or someone in my household does"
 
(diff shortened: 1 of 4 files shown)


spec/support/fill_in_the_form_steps.rb
```diff
@@ -127,7 +127,7 @@ def and_is_worried_about_mental_health
   def and_is_not_able_to_leave_home_if_absolutely_necessary
     expect(page).to have_content(I18n.t("coronavirus_form.groups.leave_home.questions.able_to_leave.title"))
 
-    choose "I cannot go out because I have coronavirus, or someone in my household does"
+    choose "I should not leave home because I have coronavirus symptoms, or someone in my household does"
```
leenagupte, line 130: Do you think it would be easier if we just pulled in the content directly from the I18n file, like we do for the expectation directly above?

issyl0: Yes, absolutely. But that's a bigger refactor to everything that we should write a separate card for.

leenagupte: I'll add a card

leenagupte: https://trello.com/c/aZdkhCC0

PR #125, opened 2020-04-17
Skip deleted questions

brucebolt:
We recently removed a question.  Since some users had already begun their journey through the form, their `questions_to_ask` session array contained the removed question's key and they were redirected to that path, even though it no longer exists.

This performs an intersection of questions the user should be asked, and the questions available in the app each time the next or previous question is being determined.  That will prevent redirection to non-existent routes.

### DIFF
--- a/spec/helpers/questions_helper_spec.rb
+++ b/spec/helpers/questions_helper_spec.rb
@@ -1,6 +1,6 @@
 RSpec.describe QuestionsHelper, type: :helper do
   before do
-    allow(helper).to receive(:questions_to_ask).and_return(%w(question_1 question_2))
+    allow(helper).to receive(:session_questions).and_return(%w(get_food afford_food))
   end
 
   describe "#determine_user_questions" do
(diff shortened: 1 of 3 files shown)


spec/helpers/questions_helper_spec.rb
```diff
@@ -1,6 +1,7 @@
 RSpec.describe QuestionsHelper, type: :helper do
   before do
-    allow(helper).to receive(:questions_to_ask).and_return(%w(question_1 question_2))
+    allow(helper).to receive(:session_questions).and_return(%w(question_1 question_2))
+    allow(helper).to receive(:all_questions).and_return(%w(question_1 question_2))
```
bilbof, line 4: By stubbing this, the [`all_questions`](https://github.com/alphagov/govuk-coronavirus-find-support/pull/125/files#diff-0890587450fdcfbd522211e4d850c710R57) method isn't tested. Perhaps we could use real questions?

Alternatively we could stub `I18n.t("coronavirus_form.groups")`. Perhaps a before step like this (I'm not sure it'll work):

```ruby
   en = { coronavirus_form: { groups: [] } }
   I18n.backend.store_translations(:en, en)
```

brucebolt: Good point 👍.  I've updated to use real questions, which means `all_quesions` is no longer stubbed.

PR #129, opened 2020-04-20
Make the tests use internationalization text rather than hardcoding questions and answers

leenagupte:
Trello: https://trello.com/c/HgWQqjjb

When we [change a question](https://github.com/alphagov/govuk-coronavirus-find-support/pull/122/commits/a02d719719717d0606046416fc1e6bce99d7ee9f), the tests fail because they [hardcode a lot of the question and answer text](https://github.com/alphagov/govuk-coronavirus-find-support/pull/122/commits/2876ee7abcf1e57dc25002e1879329830e05c0d0).

We should make the tests use the values from the internationalizations so that we don't have to do this manual dev task.

### DIFF
--- a/app/controllers/coronavirus_form/data_export_controller.rb
+++ b/app/controllers/coronavirus_form/data_export_controller.rb
@@ -43,7 +43,7 @@ class CoronavirusForm::DataExportController < ApplicationController
 private
 
   def produce_csv(results)
-    csv_data = CSV.generate do |csv|
+    csv_data = CSV.generate(col_sep: "|") do |csv|
       csv << %w(question answer date count)
       results.each do |question, value|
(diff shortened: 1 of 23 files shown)


app/controllers/coronavirus_form/data_export_controller.rb
```diff
@@ -43,7 +43,7 @@ def usage_statistics(start_date, end_date)
 private
 
   def produce_csv(results)
-    csv_data = CSV.generate do |csv|
+    csv_data = CSV.generate(col_sep: "|") do |csv|
```
issyl0, line 46: We will need to make the downstream consumers of this data (whoever they are?) aware of this.

PR #131, opened 2020-04-20
Restructure employment questions

brucebolt:
Makes changes to the user flow through the employment questions:
- "Are you self employed" is now the first question in the group
- Later questions in the group ("Have you been told to stop working" and "Are you off work") are not asked if the user is self employed
- Final question ("Are you off work") is not asked if the user has been told to stop working

Trello card: https://trello.com/c/y3mCWrG1

### DIFF
--- a/app/helpers/questions_helper.rb
+++ b/app/helpers/questions_helper.rb
@@ -56,4 +56,8 @@ module QuestionsHelper
   def all_questions
     I18n.t("coronavirus_form.groups").map { |_, group| group[:questions].keys if group[:title] }.compact.flatten.map(&:to_s)
   end
+
+  def remove_questions(questions)
+    questions_to_ask - questions
+  end
 end
--- a/app/helpers/questions_helper.rb
+++ b/app/helpers/questions_helper.rb
@@ -60,4 +60,13 @@ module QuestionsHelper
   def remove_questions(questions)
     questions_to_ask - questions
   end
+
+  def add_questions(questions, after_question)
+    index = questions_to_ask.index(after_question) + 1
+    questions_to_ask
+      .dup
+      .insert(index, questions)
+      .flatten
+      .uniq
+  end
 end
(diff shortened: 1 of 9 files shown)


app/helpers/questions_helper.rb
```diff
@@ -56,4 +56,12 @@ def last_question_seen?
   def all_questions
     I18n.t("coronavirus_form.groups").map { |_, group| group[:questions].keys if group[:title] }.compact.flatten.map(&:to_s)
   end
+
+  def remove_questions(questions)
+    updated_questions_to_ask = questions_to_ask
+    questions.each do |question|
+      updated_questions_to_ask.delete(question)
+    end
+    updated_questions_to_ask
```
bilbof, line 65: Are `questions` and `questions_to_ask` both arrays? If so, you could do `questions_to_ask - questions` here.

app/helpers/questions_helper.rb
```diff
@@ -64,4 +64,13 @@ def remove_questions(questions)
...
     updated_questions_to_ask
   end
+
+  def add_questions(questions, after_question)
+    updated_questions_to_ask = questions_to_ask
+    questions.each do |question|
+      updated_questions_to_ask.insert(updated_questions_to_ask.index(after_question) + 1, question) unless updated_questions_to_ask.include?(question)
+      after_question = question
+    end
+    updated_questions_to_ask
```
bilbof, line 74: Would it be possible to do something like this here?

```ruby
qs = questions_to_ask
index = questions_to_ask.index(after_question)
questions_to_ask.slice(0..index) + questions + questions_to_ask.slice(index..)
```

E.g.

```ruby
alphabet = ('a'..'g').to_a
alphabet.slice(0,3) + %w(x y z) + alphabet.slice(3..)
# => ["a", "b", "c", "x", "y", "z", "d", "e", "f", "g"]
```

issyl0: Having never seen that syntax before, it's quite hard to understand and would need a couple of comments as to what it's doing, I think.

app/helpers/questions_helper.rb
brucebolt: Nice one!  I didn't spot this.

app/helpers/questions_helper.rb
brucebolt: Realised I can condense this down into:
```  def add_questions(questions, after_question)
    index = questions_to_ask.index(after_question) + 1
    questions_to_ask
      .dup
      .insert(index, questions)
      .flatten
      .uniq
  end```

The `dup` is required so we don't modify the value of `questions_to_ask` since `insert` would otherwise change it's value.

PR #132, opened 2020-04-20
Bugfix: Answering "I'm not sure" should redirect users to the first question

bilbof:
When a user has all questions to answer (eg when they answer
'I'm not sure' to the question what help do you need),
we should set their questions_to_answer to be all_questions.

Currently we set the questions_to_answer to be empty, which
means users are redirected to a nil path next_quest(nil).

Please let me know if there's a better way to fix this; I'm
not familiar with this bit.

Resolves https://sentry.io/organizations/govuk/issues/1621299615
https://trello.com/c/Pw8wJYuW/288

### DIFF
--- a/app/helpers/questions_helper.rb
+++ b/app/helpers/questions_helper.rb
@@ -30,7 +30,7 @@ module QuestionsHelper
   end
 
   def questions_to_ask
-    session_questions & all_questions.map(&:to_s)
+    session_questions & all_questions
   end
 
   def session_questions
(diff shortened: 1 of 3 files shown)


brucebolt:
In `determine_user_questions`, there's a condition `if groups.empty?` which should be adding all possible questions to the array if none are selected.

brucebolt:
Looks like `all_questions` is returning an array of symbols, whereas it used to be an array of strings.  You could update `determine_user_questions` to use `all_questions.map(&:to_s)` instead of raw `all_questions`?

bilbof:
> In `determine_user_questions`, there's a condition `if groups.empty?` which should be adding all possible questions to the array if none are selected.

@brucebolt should `determine_user_questions` be called by the `need_help_with` controller? The error is occurring as a result of `questions_to_ask.first`, since `questions_to_ask` is empty.

bilbof:
Thanks for explaining @brucebolt that makes sense. I've made the change you suggested, that looks better!

PR #135, opened 2020-04-20
Fix typeset quote direction

brucebolt:
Some of the quote marks in contractions are facing the wrong direction.  Updating to keep these consistent throughout the service.


brucebolt:
I only noticed these on Windows, where the two types of quote are rendered much more distinctly than on a Mac (where they look almost identical in the default fixed width font).

PR #138, opened 2020-04-21
Javascript enabled feature specs

JonathanHallam:
- Enables javascript for feature spec. 
- Fix a bug that was introduced by freezing time in a previous PR
- Run smoke tests against staging and production in pipeline

Trello - https://trello.com/c/JWsX6enf/201-ensure-javascript-is-enabled-for-feature-tests

pipeline with smoke tests


concourse/tasks/run-smoke-tests.yml
```diff
@@ -0,0 +1,15 @@
+platform: linux
+image_resource:
+  type: docker-image
+  source:
+    repository: ((readonly_private_ecr_repo_url))
+    tag: govuk-coronavirus-find-support-feature-tests
+params:
+  TEST_URL:
+  RAILS_ENV: smoke_test
```
bilbof, line 9: One reason to have the `RAILS_ENV` be `smoke_test` rather than `test` was because we didn't want to setup a test db locally in the docker container. I wonder if this is necessary here, since in this commit it looks like active record is still loaded. Does running `TEST_URL=x RAILS_ENV=smoke_test bundle exec rspec spec/features` locally work when the DB is not available?

concourse/pipeline.yml
```diff
@@ -148,6 +152,9 @@ jobs:
...
         timeout: 5m
         params:
-          # TODO: this should point at the production URL
           URL: 'https://govuk-coronavirus-find-support-prod.cloudapps.digital/urgent-medical-help'
-          MESSAGE: "Checks that the application deployed to production is not serving HTTP error codes. If this fails, you should investigate immediately."
+          MESSAGE: "Checks that the application deployed to production is not serving HTTP error codes"
+      - task: run-smoke-tests
+        file: git-master/concourse/tasks/run-smoke-tests.yml
+        params:
+          TEST_URL: 'https://find-coronavirus-support.service.gov.uk/urgent-medical-help'
```
bilbof, line 160:
```suggestion
          TEST_URL: 'https://find-coronavirus-support.service.gov.uk'
```

concourse/tasks/run-smoke-tests.yml
Rosa-Fox: @bilbof ah yeah... it does not as the only available databases configurations are: `default, development, test, production`. 

Was the reason for not wanting to set up the test db in the docker container more to do with needing to get something up and running quickly, or was it to do with not wanting to persist the data/needing to identify that the data is from the smoke test to then remove it (for prod)?

If it is ok to persist, then I guess we can remove the `smoke_test` rails env and use `test`... though our docker-image for `run-smoke-tests`  would need to support chrome and postgres... so I guess we could do something to `run-quality-checks` and load that same Dockerfile but just run `bundle exec rspec spec/features` with the `test_url` as opposed to `rake`? May be good to catch up about this on a quick call?

concourse/pipeline.yml
Rosa-Fox: Thanks!

concourse/tasks/run-smoke-tests.yml
bilbof: Yeah, iirc when we set this up we didn't have the docker container. Now that we do, perhaps we could switch to using the test env. We shouldn't need a DB connection to run the smoke tests, but it won't hurt.

>  could do something similar to run-quality-checks and load that same Dockerfile but just run bundle exec rspec spec/features with the test_url as opposed to rake

Yeah I think that makes sense, I'd go for that. I'm available for a call if you want to discuss it.

spec/controllers/coronavirus_form/able_to_leave_controller_spec.rb
```diff
@@ -8,6 +8,10 @@
       Timecop.freeze(Time.utc(2020, 3, 1, 10, 43, 45))
     end
 
+    after do
```
huwd, line 11: great catch!

concourse/pipeline.yml
```diff
@@ -148,6 +152,9 @@ jobs:
         file: git-master/concourse/tasks/smoke-test.yml
         timeout: 5m
         params:
-          # TODO: this should point at the production URL
           URL: 'https://govuk-coronavirus-find-support-prod.cloudapps.digital/urgent-medical-help'
-          MESSAGE: "Checks that the application deployed to production is not serving HTTP error codes. If this fails, you should investigate immediately."
+          MESSAGE: "Checks that the application deployed to production is not serving HTTP error codes"
+      - task: run-smoke-tests
+        file: git-master/concourse/tasks/run-smoke-tests.yml
```
huwd, line 158: :thinking: so we're running this against both staging and prod...
I'm a little confused about what we'd expect to be stored in the postgres DB in the image and what might actually end up in staging / prod's database.

Why won't these tests incriment the production postgres counters when you hit the results page?
Isn't there a risk we're inflating real numbers every time we run the tests?

bilbof: In the business form smoke tests we don't write to the DB, and rely on user activity to alert us of failures:

https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/master/app/controllers/coronavirus_form/check_answers_controller.rb#L12

Rosa-Fox: thanks, will add something similar

app/controllers/coronavirus_form/able_to_leave_controller.rb
```diff
@@ -35,11 +35,17 @@ def time_hour_floor
...
+      )
+    end
+  end
 
-    FormResponse.create(
-      form_response: redacted_session,
-      created_at: time_hour_floor,
-    )
+  def smoke_tester?
+    smoke_test_header = request.env["HTTP_SMOKE_TEST"]
```
huwd, line 47: nice!

concourse/tasks/run-smoke-tests.yml
Rosa-Fox: added a commit to use the header to check if smoke test.

bilbof:
👏 nice work

PR #150, opened 2020-04-23
Remove urgent medical help question

injms:
What
----

We need to remove the question asking whether a user requires urgent medical help, as this has been found to be confusing.

~This should not be merged in until the [dependent card](https://trello.com/c/r1ySyxao) has been completed.~ Dependencies have been completed ✅ 

## Review

❓ I've removed some tests and rejigged others - are there any extra tests that should be added?
❓ I'm not confident with the feature test changes - is there anything I've missed?
❓ Should the escape link appear on all questions?

Links
-----

Trello card 👉  https://trello.com/c/Ww64j5Xy


huwd:
[This card is in newly done](https://trello.com/c/r1ySyxao/324-add-if-you-need-medical-support-section-into-start-page), so marking this as ready to merge

app/helpers/questions_helper.rb
```diff
@@ -46,7 +46,7 @@ def last_question
   end
 
   def first_question_seen?
-    session[:urgent_medical_help].present?
+    session[:need_help_with].present?
```
brucebolt, line 49: The `first_question_seen?` method can be completely removed, since it was only used in `ApplicationController`, but is no longer being called.

injms: I've removed this - thanks Bruce!

brucebolt:
I think a redirect from `/urgent-medical-help` to `/need-help-with` would be a good idea, to maintain any bookmarks/links from elsewhere.

Something like this in routes should work:
```get "/urgent-medical-help", to: redirect("/need_help_with")```

File: config/routes.rb [hunk 2]
```diff
@@ -11,6 +11,10 @@
     get "/cookies", to: "cookies#show"
     get "/accessibility-statement", to: "accessibility_statement#show"
 
+    # Redirect for deleted question and page (301 is default)
+    get "/urgent-medical-help", to: redirect("/need_help_with")
+    get "/get-help-from-nhs", to: redirect("/need_help_with")
```
brucebolt, line 16:
```suggestion
    get "/get-help-from-nhs", to: redirect("/need-help-with")
```
Looks like there's a small typo in the redirect URL (I typed in wrong in my last comment 🤦 ).

injms: That'll teach me to copy-and-paste without checking! Thanks Bruce.

brucebolt: There's also the same typo on the line above this one

injms: 🤦 x 10

PR #151, opened 2020-04-23
Add basic auth to data export endpoint

bilbof:
What
----

This requires basic authentication when accessing the `/data-export.csv` path.

Though no personal information is provided via this endpoint, end users don't need access to this reporting tool.

The relevant secrets have already been added. I'll need to share the username/password with those who require access. Should the requirements of the reporting tool change or if we need to share these credentials more widely, we should look at using signon for authentication.

Links
-----

https://trello.com/c/KEUQdHPk/328-put-data-export-endpoint-behind-basic-auth


huwd:
I'm gonna merge this one

PR #155, opened 2020-04-24
Tracking no results page

Rosa-Fox:
What
----

- Changes results page title depending on wether there are any specific results displayed to the user
- This will allow analysts to determine how many users were not presented with specific results
- Adds event tracking to the ‘Start again’ link

FYI

This solution is a quick way to implement adding event tracking. Ian is going to look into a more GOV.UK like solution which will be easier to maintain than keeping track of changes inline to elements, but we think this is ok for our current need.

How to review
-------------
Harriet (analyst) has already confirmed the naming for the `Event category`, `Action` and `Label`.
@injms and @huwd and I have tested it manually using Google Analytics Debugger. To test we:

- Accepted cookies
- Navigated to the results page
- Used the Google Analytics Debugger extension to see the event fired when the 'Start Again' link is clicked
- Denied cookies
- Checked that the event did not fire 

Links
-----

[Trello](https://trello.com/c/GTMlUZ8H/195-add-google-analytics-events-tracking-to-the-validation-errors-and-no-results-page)

### DIFF
--- a/app/views/coronavirus_form/results.html.erb
+++ b/app/views/coronavirus_form/results.html.erb
@@ -15,7 +15,7 @@
     margin_top: 0,
     margin_bottom: 6,
   } %>
-  <%= link_to t("coronavirus_form.results.header.start_again_text"), clear_session_path, class: "govuk-link" %>
+  <%= link_to t("coronavirus_form.results.header.start_again_text"), clear_session_path, class: "govuk-link", onclick: "ga('send', 'event', 'StartAgain', 'StartAgainClicked', '/results');" %>
 <% end %>
 
 <% if result_groups(session).empty? %>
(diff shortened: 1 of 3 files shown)


app/views/coronavirus_form/results.html.erb
```diff
@@ -1,4 +1,8 @@
-<% content_for :title do %><%= t('coronavirus_form.results.header.title') %><% end %>
+<% if result_groups(session).empty? %>
+  <% content_for :title do %><%= t('coronavirus_form.results.header.title_no_results') %><% end %>
```
huwd, line 2: @injms pedantic question, do we prefer flow control to be in a helper or we cool with if statements in the view

Rosa-Fox: @huwd I will add a commit for this as. I sort of also realised I need to change the title for the meta_tags content below as well... so helper methods will stop the logic in the views building up too much.

Rosa-Fox: Added a commit for this.

huwd:
Looks good!

PR #156, opened 2020-04-24
Add enhanced ecommerce to results page

injms:
What
----

Add enhanced ecommerce to help the analytics reporting for this service. 

The results page for this service is different depending on what a user needs help with - because of this we can't tell what links are being seen the most, and what links are being clicked the most. The ecommerce plugin will help us do that by monitoring what positions the different links appear, and will help us to see what links are most useful for users. We can then analyse this information to improve the quality of the results.

How to review
-------------

Use a private mode in your browser.

Check that Google Analytics is not being blocked by your browser

1. Go to the service
1. Accept cookies
1. Go through service
1. On results page:
   1. Check that the page view is being fired with all the impressions of the results
   1. Check that when a link is clicked it is firing a GA event

Close the browser completely.

In another private browser window:

1. Go to the service
1. Don't accept cookies
1. Go through the service
1. On the results page:
   1. Check that the page view isn't fired at all
   1. Check that no events are fired when a link is clicked

Links
-----

Trello card 👉 https://trello.com/c/mRtFakg3

### DIFF
--- a/app/views/coronavirus_form/results.html.erb
+++ b/app/views/coronavirus_form/results.html.erb
@@ -22,12 +22,14 @@
 <% if result_groups(session).empty? %>
   <%= sanitize(t("coronavirus_form.results.no_results")) %>
 <% else %>
-  <% result_groups(session).each do |group| %>
-    <%= render "components/actions-group", {
-      title: group[1][:heading],
-      subsections: group[1][:questions]
-    } %>
-  <% end %>
+  <div data-track-ec-list="Find coronavirus support">
+    <% result_groups(session).each do |group| %>
+      <%= render "components/actions-group", {
+        title: group[1][:heading],
+        subsections: group[1][:questions]
+      } %>
+    <% end %>
+  </div>
 <% end %>
 
 <%= render "components/callout", {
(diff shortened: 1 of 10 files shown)


app/views/coronavirus_form/results.html.erb
```diff
@@ -17,12 +17,14 @@
 <% if result_groups(session).empty? %>
   <%= sanitize(t("coronavirus_form.results.no_results")) %>
 <% else %>
-  <% result_groups(session).each do |group| %>
-    <%= render "components/actions-group", {
-      title: group[1][:heading],
-      subsections: group[1][:questions]
-    } %>
-  <% end %>
+  <div data-track-ec-list="Find coronavirus support">
```
alex-ju, line 20: Non-blocker. If time allows it would be good to avoid this extra div and pass the data attribute to the action-group component; as a suggestion, calling it with `data_attributes: { "track-ec-list": "Find coronavirus support" }` and updating the component to have [`data_attributes` in the section element](https://github.com/alphagov/govuk-coronavirus-find-support/blob/master/app/views/components/_actions-group.html.erb#L5).

injms: We only want the `data-track-ec-list` attribute to be added to the one element wrapping all of the results so the position of each result can be reported.

Because the action-group component is called multiple times on the results page this would cause `data-track-ec-list` to be added to multiple sections on the page, rather than once. 

This would mean that the position of each result would be reported in relation to each section, rather than in relation to the entire page.

PR #157, opened 2020-04-25
Change self employment flow

brucebolt:
What
----

No longer skips the employment questions if a user is unsure about being self employed.

Previously they would have skipped the following two questions.

How to review
-------------

- Review the code.
- Run the app locally.  Select you need help with 'Unemployment'.  Check whether selecting the different options on 'Are you self employed' question skips the next two questions.

Links
-----

Trello card: https://trello.com/c/dbly3cUo

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -274,7 +274,6 @@ en:
                 label: "Not sure"
             skip_next_question_options:
               - "Yes"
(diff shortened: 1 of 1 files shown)


brucebolt:
Yeah, I only saw Hannah's question after creating the PR myself

PR #164, opened 2020-04-30
Fix typo in cookie name

connordoner:
The name of the cookie used to store the session ID is `_session_id` yet is referred to on the cookies page as `_sessions_id`.


connordoner:
> Good spot 👍. Thank you for your contribution @connordoner.

No problem, @brucebolt - thanks for all of your hard work on this! Very encouraging to see how quickly GOV.UK has adjusted to coping with these new challenges in a way that Directgov and the dreaded Stellent could never have... Martha would be proud. :grin:

PR #175, opened 2020-05-07
Add devolved links to results page

JonathanHallam:
What
----

This adds in the functionality to filter our results based on what country the user has said that they are in. In order to make this work we've had to do some funny things with some of the tests and I think ideally we'd make further changes to decouple our testing of code from the copy we have.

## New question
Screenshot 2020-05-07 at 08 56 41

Trello - https://trello.com/c/sW6sLedW/329-add-devolved-links-to-the-results-page

### DIFF
--- a/app/helpers/results_helper.rb
+++ b/app/helpers/results_helper.rb
@@ -23,7 +23,7 @@ module ResultsHelper
 
   def filter_results_by_nation(question_results)
     question_results[:items] = question_results[:items].select do |item|
-      item[:show_to_nations].nil? || item[:show_to_nations].include?(session[:where_live])
+      item[:show_to_nations].nil? || item[:show_to_nations].include?(session[:nation])
     end
     question_results
   end
--- a/app/helpers/results_helper.rb
+++ b/app/helpers/results_helper.rb
@@ -13,11 +13,19 @@ module ResultsHelper
   end
 
   def filter_questions_by_session(group_key, session)
-    I18n.t("results_link.#{group_key}").each_with_object([]) do |question, array|
+    group = I18n.t("results_link.#{group_key}").dup
+    group.each_with_object([]) do |question, array|
       if question[1][:show_options].include?(session[question[0]])
-        array << I18n.t("results_link.#{group_key}.#{question[0]}")
+        array << filter_results_by_nation(I18n.t("results_link.#{group_key}.#{question[0]}").dup)
       end
     end
+  end
+
+  def filter_results_by_nation(question_results)
+    question_results[:items] = question_results[:items].select do |item|
+      item[:show_to_nations].nil? || item[:show_to_nations].include?(session[:where_live])
+    end
+    question_results
   end
 
   def relevant_group_keys
@@ -27,7 +35,7 @@ module ResultsHelper
     # :help and :filter_questions which are not really groups
     # empty.
     if selected_groups.blank?
-      return I18n.t("coronavirus_form.groups").keys - %i[filter_questions help leave_home]
+      return I18n.t("coronavirus_form.groups").keys - %i[filter_questions help leave_home location]
     end
 
     selected_groups # Otherwise we can use selected groups
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -103,9 +103,9 @@ en:
   session_expired:
     title: Your session has ended due to inactivity
     body_text: |
-      <p class="govuk-body">Your session has ended because you have not done anything for 4 hours. You'll have to start again.</p>
+      <p class="govuk-body">Your session has ended because you have not done anything for 4 hours. You’ll have to start again.</p>
 
-      <p class="govuk-body">We do this for your security. We've deleted all the details you entered to protect your data.</p>
+      <p class="govuk-body">We do this for your security. We’ve deleted all the details you entered to protect your data.</p>
 
       <h2 class="govuk-heading-m">If you didn’t expect to see this page</h2>
       <p class="govuk-body">You may be seeing this page because you’ve turned off cookies in your browser. You’ll need to <a class="govuk-link" href="https://ico.org.uk/your-data-matters/online/cookies/" rel="external" target="_blank">turn cookies on</a> before you can use this service.</p>
@@ -145,7 +145,7 @@ en:
           text_html: These cookies are required for this service to operate. We do not need to ask permission to use them.
           cookies:
             - - text: cookie_preferences_set
-              - text: Lets us know whether you've already set your cookies preferences.
+              - text: Lets us know whether you’ve already set your cookies preferences.
               - text: 1 year
             - - text: cookie_preferences
               - text: Let us know what your cookie preferences are.
@@ -202,7 +202,6 @@ en:
                 label: "Wales"
               optopn_northern_ireland:
                 label: "Northern Ireland"
-            custom_select_error: "Select where you live"
       filter_questions:
         questions:
           need_help_with:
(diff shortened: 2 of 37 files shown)


config/routes.rb
```diff
@@ -12,8 +12,12 @@
     get "/accessibility-statement", to: "accessibility_statement#show"
 
     # Redirect for deleted question and page (301 is default)
-    get "/urgent-medical-help", to: redirect("/need-help-with")
-    get "/get-help-from-nhs", to: redirect("/need-help-with")
+    get "/urgent-medical-help", to: redirect("/where-live")
+    get "/get-help-from-nhs", to: redirect("/where-live")
+
+    # Question: Where do you live
+    get "/where-live", to: "where_live#show"
```
brucebolt, line 19: I wonder if the route should just be a single word (like `location`) rather than part of a sentence?

config/routes.rb
```diff
@@ -12,8 +12,12 @@
     get "/accessibility-statement", to: "accessibility_statement#show"
 
     # Redirect for deleted question and page (301 is default)
-    get "/urgent-medical-help", to: redirect("/need-help-with")
-    get "/get-help-from-nhs", to: redirect("/need-help-with")
+    get "/urgent-medical-help", to: redirect("/where-live")
+    get "/get-help-from-nhs", to: redirect("/where-live")
```
brucebolt, line 16: Could the first question be defined as an application wide constant, which can then be referenced elsewhere, e.g. https://github.com/alphagov/govuk-coronavirus-find-support/pull/175/commits/ba03935f826cfa74c3bdaf62ae6ea753369d7ce7#diff-55c5b7aecfb519d0e4880eaf2788eb6eR40 and all the tests?  It will make it easier to change the first question if we need to do this again.

config/locales/en.yml
```diff
@@ -404,22 +403,34 @@ en:
           - text: 'Find helplines <a href="https://www.gov.uk/government/publications/coronavirus-covid-19-and-domestic-abuse/coronavirus-covid-19-support-for-victims-of-domestic-abuse" class="govuk-link">if you’re a victim of domestic abuse or feel at risk of abuse</a>, or if you are a <a href="https://www.gov.uk/guidance/coronavirus-covid-19-victim-and-witness-services" class="govuk-link">victim or witness of a crime</a> (these pages have no quick escape)'
           - text: "Get safety advice for survivors (Women’s Aid)"
             href: "https://www.womensaid.org.uk/covid-19-coronavirus-safety-advice-for-survivors/"
-          - text: "Get help if you’re in Northern Ireland (Nexus NI)"
+          - text: "Get help from Nexus NI"
             href: "https://nexusni.org/domestic-and-sexual-abuse-24-hour-helpline/"
-          - text: "Get help if you live in Wales (Live Fear Free)"
+            show_options:
```
brucebolt, line 408: I'm thinking `show_options` doesn't necessarily explain what this value is doing.  Maybe `show_to_nations` or `only_show_to_nations` or something else might be a better choice of name?

app/helpers/results_helper.rb
```diff
@@ -13,21 +13,29 @@ def result_groups(session)
...
       if question[1][:show_options].include?(session[question[0]])
-        array << I18n.t("results_link.#{group_key}.#{question[0]}")
+        array << filter_results_by_country(I18n.t("results_link.#{group_key}.#{question[0]}").dup)
       end
     end
   end
 
+  def filter_results_by_country(question_results)
+    question_results[:items] = question_results[:items].select do |item|
+      item[:show_options] == nil || item[:show_options].include?(session[:where_live])
```
brucebolt, line 26:
```suggestion
      item[:show_options].nil? || item[:show_options].include?(session[:where_live])
```

app/helpers/results_helper.rb
```diff
@@ -13,21 +13,29 @@ def result_groups(session)
...
+    group = I18n.t("results_link.#{group_key}").dup
+    group.each_with_object([]) do |question, array|
       if question[1][:show_options].include?(session[question[0]])
-        array << I18n.t("results_link.#{group_key}.#{question[0]}")
+        array << filter_results_by_country(I18n.t("results_link.#{group_key}.#{question[0]}").dup)
       end
     end
   end
 
+  def filter_results_by_country(question_results)
```
brucebolt, line 24: A little on the nitpick side, but I think `nation` is a better word to use than `country` (and seems to be the word we use on GOV.UK)

app/helpers/results_helper.rb
```diff
@@ -13,21 +13,29 @@ def result_groups(session)
   end
 
   def filter_questions_by_session(group_key, session)
-    I18n.t("results_link.#{group_key}").each_with_object([]) do |question, array|
+    group = I18n.t("results_link.#{group_key}").dup
+    group.each_with_object([]) do |question, array|
       if question[1][:show_options].include?(session[question[0]])
-        array << I18n.t("results_link.#{group_key}.#{question[0]}")
+        array << filter_results_by_country(I18n.t("results_link.#{group_key}.#{question[0]}").dup)
```
brucebolt, line 19: Is the duplication (`.dup`) of the hash necessary here and on line 16?  It seems to make no difference when I remove them and run the application locally.

config/routes.rb
JonathanHallam: So i've called the group that this question is in location just in case we need to add any more granularity. How would you feel about calling this "country"?

JonathanHallam: just seen later comments, how do you feel about "nation"? :D

config/routes.rb
JonathanHallam: yeah I think this is a good shout. For the sake of getting this released can we add it as a new story though? Kind of aware that this has taken me a while to get in and it'd be good to get this functionality added.

config/routes.rb
JonathanHallam: Just seen later comments, how do you feel about "nation"? :D

config/locales/en.yml
JonathanHallam: done 👍

app/helpers/results_helper.rb
JonathanHallam: done 👍 thanks :)

app/helpers/results_helper.rb
JonathanHallam: done, removed it from method names and commit messages

config/routes.rb
brucebolt: Yes, that's absolutely fine with me 👍

config/routes.rb
brucebolt: Nation is good.  I just wanted to avoid any potential controversy over the country status of the four constituents of the UK.

config/locales/en.yml
```diff
@@ -418,20 +444,48 @@ en:
             href: "https://www.gov.uk/government/news/government-agrees-measures-with-energy-industry-to-support-vulnerable-people-through-covid-19"
```
hannahjump, line 444: Remove - this is now a very old news story

config/locales/en.yml
```diff
@@ -418,20 +444,48 @@ en:
...
+              - England
+          - text: "What to do if you cannot pay your rent or mortgage (Shelter Scotland)"
+            href: "https://scotland.shelter.org.uk/get_advice/scottish_housing_advice_coronavirus_COVID_19"
+            show_to_nations:
+              - Scotland
+          - text: "What to do if you cannot pay your rent or mortgage (Shelter Cymru)"
+            href: "https://sheltercymru.org.uk/get-advice/coronavirus/"
+            show_to_nations:
+              - Wales
           - text: "What to do if you cannot pay your rent, mortgage or bills (Citizens Advice)"
```
hannahjump, line 457: Change to "What to do if you cannot pay your bills or mortgage (Citizens Advice)"

Reason - little information here about rent

config/locales/en.yml
```diff
@@ -418,20 +444,48 @@ en:
...
             href: "https://www.citizensadvice.org.uk/debt-and-money/if-you-cant-pay-your-bills-because-of-coronavirus/"
+            show_to_nations:
+              - England
+          - text: "What to do if you cannot pay your rent, mortgage or bills (Citizens Advice Scotland)"
```
hannahjump, line 461: Change to "What to do if you cannot pay your bills or mortgage (Citizens Advice Scotland)"

Reason - little information here about rent

config/locales/en.yml
```diff
@@ -418,20 +444,48 @@ en:
...
+            href: "https://www.citizensadvice.org.uk/scotland/debt-and-money/if-you-cant-pay-your-bills-because-of-coronavirus/"
+            show_to_nations:
+              - Scotland
+          - text: "What to do if you cannot pay your rent, mortgage or bills (Citizens Advice)"
```
hannahjump, line 465: Change to "What to do if you cannot pay your bills or mortgage (Citizens Advice)"

Reason - little information here about rent

config/locales/en.yml
JonathanHallam: 👍

config/locales/en.yml
JonathanHallam: 👍

config/locales/en.yml
JonathanHallam: 👍

config/locales/en.yml
JonathanHallam: 👍

PR #176, opened 2020-05-07
Change label for "Feeling unsafe" on filter question

brucebolt:
What
----

We currently use the group labels as checkbox options in the "What do you need help" filter question.  For "Feeling unsafe" we want to have a different label for the checkbox in this question, but retain the existing group title.

Therefore adding a `need_help_with_label` value to the group.  This will also us to override the label that is displayed in the "What do you need help with" question for a specific group.

Screenshot 2020-05-07 at 10 29 09

How to review
-------------

- Review code
- Run application locally and navigate to `/need-help-with`

Links
-----

Trello card: https://trello.com/c/NyFF0UUN

### DIFF
--- a/app/controllers/coronavirus_form/need_help_with_controller.rb
+++ b/app/controllers/coronavirus_form/need_help_with_controller.rb
@@ -30,7 +30,7 @@ private
     groups_items = groups_hash.map do |key, title|
       {
         value: title,
-        label: title,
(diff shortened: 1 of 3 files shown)


bilbof:
@brucebolt following https://github.com/alphagov/govuk-coronavirus-find-support/pull/177, I think if you rebase the tests would pass

PR #185, opened 2020-05-11
Set up usage statistics exports to a google sheet

huwd:
What
----

We want a way for PAs to get a hold of the usage data automatically.
Initially we tried to do this by removing the basic auth export endpoint and switch over to a rake task that exports and posts data to a sheet. However this has proved to be difficult due to some cross domain issues and complexity around Service Accounts and Auth tokens.

A simplier way has been to allow google sheets to scrape data AND authenticate with the existing basic auth.

To do this we have the following changes in the app:

- Changes Data export endpoint to export all time if no explicit dates are provided
- Add docs on how this works to docs repo
- Add further test for 401 if basic auth is not provided

And the following changes outside of the app:
- Write a custom JS function in a google sheet that can scrape CSV through basic auth
- Write a sheets UI widget so a PA can manually trigger a refresh
- Set google triggers up so it will also run a refresh automatically between midnight and 1am.

How to review
-------------

- Read over the code here and look for errors
- Check the test suite runs on CI
- Read over the docs added in this PR and ensure they are clear enough.
- Following the links in the docs:
  - Check you can get access to the sheet.
  - Perform a manual refresh in the UI (instructions should be clear in the docs, shout if not!)
  - If you are a JS person, you can also review the JS code in Google Scripts linked to in the docs.

If you can't get access to any of the google files, raise that as a PR comment and I'll ensure permissions are revised.

Links
-----

Trello: https://trello.com/c/YdlVTdIo/346-export-aggregate-data-to-a-google-sheet


docs/usage_statistics_exports.md
```diff
@@ -0,0 +1,25 @@
+# How can you export data
+
+Aggregated anonymous daily data can be extracted from the application for any
+specified time period using the `/data-export` route.  Output is available in
+CSV format.
+
+For example, to get data for a five day period starting 10th April 2020:
+
+     <path-to-service>/data-export.csv?start_date=2020-04-08&end_date=2020-04-15
```
brucebolt, line 9: I know you've just copied this from the README, but the date range gives seven days rather than five days (as mentioned above)

docs/usage_statistics_exports.md
```diff
@@ -0,0 +1,25 @@
...
+
+If a date is not supplied it will output a CSV covering all time periods:
```
brucebolt, line 11: It looks like the default parameters (https://github.com/alphagov/govuk-coronavirus-find-support/pull/185/files#diff-af972ce9990665f118a3e71481574974R19-R20) are from 2020-01-01 to 86400 seconds from now (i.e. 24 hours time)

docs/usage_statistics_exports.md
```diff
@@ -0,0 +1,25 @@
...
+
+
+     <path-to-service>/data-export.csv
+
+## Google App scripts & Sheets
+
+PAs have access to the aggrigate data through a google sheet which can be found here:
```
brucebolt, line 18:
```suggestion
Performance Analysts have access to the aggregate data through a Google sheet which can be found here:
```

docs/usage_statistics_exports.md
```diff
@@ -0,0 +1,25 @@
...
+[Find Support Tool - Usage Statistics](https://docs.google.com/spreadsheets/d/1d0tqlBH83TkQ8oLGDSpbIDs0YWe2burhXA-m7OR3Irk)
+
+The sheet uses a fetch script written with with Google App scripts ([see dashboard here](https://script.google.com/home/projects/120es-7Lec6yZo9VdHMigppso2eaIeh1Nq39XevJj6LTp5Vo8ySGCF59P)).
```
brucebolt, line 21:
```suggestion
The sheet uses a fetch script written with Google App scripts ([see dashboard here](https://script.google.com/home/projects/120es-7Lec6yZo9VdHMigppso2eaIeh1Nq39XevJj6LTp5Vo8ySGCF59P)).
```

spec/requests/data_export_spec.rb
```diff
@@ -68,7 +68,28 @@
       ]
     end
 
-    it "shows all expected responses in CSV format" do
+    it "shows all expected responses in CSV format for a 1000 year range if start and end date are not provided" do
```
brucebolt, line 71: It looks like the default parameters are 1 day ahead of the current date, rather than 1000 years.  Also, I can't see anything in the test that checks for the 1000 year range.

Maybe a better solution would be to export all data (with no arbitrary time limits) if no parameters are specified?

huwd: Ah great spot, sorry when I squashed the commits I thought i'd got all references to this!
I'll amend that and the fixtures too.

Yeah had tried to just remove that, but that would have required more significant changes to the where clause that exports this. I figured it'd just be quicker to accept that 1st of Jan (before the app was built) to tomorrow, effectively always equals all time.

docs/usage_statistics_exports.md
huwd: Great spot!

docs/usage_statistics_exports.md
huwd: Have amended the docs to clarify, let me know what you think?

spec/requests/data_export_spec.rb
huwd: Sorry was half way through mucking around with my git history and forgot i was supposed to be on a call. try now

huwd:
Thanks bruce, and great review those improvements are greatly appreciated!

PR #192, opened 2020-05-13
Ensure GovUK analytics hasn't been blocked before trying to call it.

steveholmes:
Previously the call was throwing exception when it had been blocked and it prevented initAll() being called.

What
----
Moved the call to GOVUK.analyticsInit to the end of the script and check loading the analytics has worked before trying to call it.

How to review
-------------
Step 1. Block analytics from loading using either a browser privacy plugin - or Brave browser.
Step 2. Load any page and ensure no javascript errors in the console log

Links
-----
https://trello.com/c/vduHcnvs/411-js-error-when-analytics-is-blocked

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -6,5 +6,7 @@
 //= require analytics-enhancedEcommerce
 //= require cookies
 window.CookieSettings.start()
-window.GOVUK.analyticsInit()
 window.GOVUKFrontend.initAll()
+if (window.GOVUK.analyticsInit) {
+  window.GOVUK.analyticsInit()
+}


issyl0:
Looks good but this branch needs a rebase, please.

The `push` CI passed, but the `pull_request` CI failed. This is because the `actions/checkout@v2` action of the `pull_request` does a merge with master before running the tests. Some RuboCop changes happened on master.

issyl0:
Also one tiny style nit: GovUK in your commit message should be GOV.UK, please. You can read [this blog post](https://gds.blog.gov.uk/2014/05/01/when-to-use-caps-around-gov-uk/) for more context!

app/assets/javascripts/application.js
```diff
@@ -6,5 +6,7 @@
 //= require analytics-enhancedEcommerce
 //= require cookies
 window.CookieSettings.start()
-window.GOVUK.analyticsInit()
 window.GOVUKFrontend.initAll()
+if (window.GOVUK.analyticsInit) {
```
injms, line 10: Can we move this if statement inside the `analyticsInit()` function please? That way we keep the `application.js` free of any logic.

steveholmes:
Urgh - realised this wasn't some super smart analytics blocking but was just blocking the analytics.js file in the dev environment based on the filename alone.
In production as the files are combined into application.js the code isn't blocked.
Therefore this fix is not needed.
TIL!

PR #195, opened 2020-05-14
Prometheus instrumentation

danwilliams:
## Changes

Replicated the instrumentation changes from the `govuk-coronavirus-vulnerable-people-form` repo here, such that request level metrics are served on the `metrics` URL. 

## How to review

A test has been added to check that the `metrics` endpoint returns a 200, this test can be run (along with the other tests in the repo) using the usual command.

```bash
bundle exec rake
```

In addition to performing a manual test one can run the server via:

```bash
foreman start
```

And go to [http://localhost/metrics](http://localhost/metrics) and see that metrics are being returned. 

**Note**: Metrics are gathered as requests are made, so on the first visit it might look somewhat empty. If you refresh the page you'll see the metrics for the first load displayed... if you visit other URLs on the app, on subsequent visits to the metrics page you'll see data for those URLs displayed too.

## Links

https://trello.com/c/yOf6sEIk/381-instrument-the-find-support-app

### DIFF
--- a/concourse/tasks/deploy-to-govuk-paas.yml
+++ b/concourse/tasks/deploy-to-govuk-paas.yml
@@ -57,3 +57,8 @@ run:
 
       cf v3-zdt-push govuk-coronavirus-find-support --wait-for-deploy-complete --no-route
       cf map-route govuk-coronavirus-find-support cloudapps.digital --hostname "$HOSTNAME"
+
+      cf map-route govuk-coronavirus-find-support find-coronavirus-support.service.gov.uk --path metrics
+      cf map-route govuk-coronavirus-find-support cloudapps.digital --hostname govuk-coronavirus-find-support-prod --path metrics
+      cf bind-route-service find-coronavirus-support.service.gov.uk re-ip-whitelist-service --path metrics
+      cf bind-route-service cloudapps.digital re-ip-whitelist-service --hostname govuk-coronavirus-find-support-prod --path metrics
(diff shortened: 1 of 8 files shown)


concourse/tasks/deploy-to-govuk-paas.yml
```diff
@@ -57,3 +57,8 @@ run:
 
       cf v3-zdt-push govuk-coronavirus-find-support --wait-for-deploy-complete --no-route
       cf map-route govuk-coronavirus-find-support cloudapps.digital --hostname "$HOSTNAME"
+
+      cf map-route govuk-coronavirus-find-support coronavirus-find-support.service.gov.uk --path metrics
```
bilbof, line 61:
```suggestion
      cf map-route govuk-coronavirus-find-support find-coronavirus-support.service.gov.uk --path metrics
```

concourse/tasks/deploy-to-govuk-paas.yml
```diff
@@ -57,3 +57,8 @@ run:
...
+      cf map-route govuk-coronavirus-find-support cloudapps.digital --hostname govuk-coronavirus-find-support-prod --path metrics
+      cf bind-route-service coronavirus-find-support.service.gov.uk re-ip-whitelist-service --path metrics
```
bilbof, line 63:
```suggestion
      cf bind-route-service find-coronavirus-support.service.gov.uk re-ip-whitelist-service --path metrics
```

PR #216, opened 2020-05-18
Rake task: Export results links as YAML

huwd:
What
----

- Adds IDs to all result links
- Adds an export rake task to easily get content out into a CSV
- Immedieatly extends the content export to support the new Support and Advice sections

Why
----

There are regular content updates to the triage tool, but we don't have a publishing interface.
Given the fiddly nature of these updates there is less chance for Human Error if we can do the updates in a google sheet compared to using a doc as current.

This PR adds an export task that takes relevant data from the locale file and uses it to produce a CSV that we can import to google sheets.

Currently that is a manual process (it saves the CSV to tmp) but this can be automated in future for a dev to run and update locally via rake and the Google API.

It has already been used to produce [this spreadsheet here](https://docs.google.com/spreadsheets/d/11jtiwMovC9736F7ZV9k3gDbOpMMGqnSsiRmlD0JFCpQ/edit#gid=0)

How to review
-------------

- Check Rubocop Passes
- Reivew the test suite and check it makes snse
- Check the tests pass
- Look over the logic and offer any comments.

Links
-----

- Trello [Results page update (3/3): Rearrange content](https://trello.com/c/rbRyihX0/401-results-page-update-3-3-rearrange-content)
- Spreadsheet - [Find Support - Results Content Sheet](https://docs.google.com/spreadsheets/d/11jtiwMovC9736F7ZV9k3gDbOpMMGqnSsiRmlD0JFCpQ/edit#gid=0)


spec/lib/content_exporter_spec.rb
```diff
@@ -0,0 +1,199 @@
+locale_results_link_fixture = {
+  group_one: {
+    subgroup_one: {
+      title: "I am the title for Group one subgroup one",
+      items: [
+        {
+          id: "0001",
+          support_and_advice: false,
+          text: "This is a row that only has text, it will appaer like a paragraph",
```
danwilliams, line 9: Here and elsewhere in the document `appear` is misspelt as `appaer`.

spec/lib/content_exporter_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+        {
+          id: "0002",
+          support_and_advice: false,
+          text: "This is a row has text and an href, it will appaer as an anchor tag",
+          href: "http://test.stubbed.gov.uk",
+        },
+        {
+          id: "0003",
+          support_and_advice: false,
+          text: "This is a row has text, an href and group criteira, it will appear as an anchor tag if the user's answers match the criteria",
```
danwilliams, line 20: Here and elsewhere in the document `criteria` is misspelt as `criteira`.

spec/lib/content_exporter_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+        subgroup_title: "I am the title for Group one subgroup one",
+        support_and_advice: false,
+        text: "This is a row that only has text, it will appaer like a paragraph" },
+      { id: "0002",
+        group_title: "I am the title for Group one",
+        href: "http://test.stubbed.gov.uk",
+        show_to_nations: "",
+        subgroup_title: "I am the title for Group one subgroup one",
+        support_and_advice: false,
+        text: "This is a row has text and an href, it will appaer as an anchor tag" },
```
danwilliams, line 71: `This is a row has text` is better phrased as `This is a row and has text`.

spec/lib/content_exporter_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+        text: "This is a row that only has text, it will appaer like a paragraph" },
+    ]
+
+    before do
+      allow(I18n).to receive(:t).with("results_link") { locale_results_link_fixture }
+      allow(I18n).to receive(:t).with("coronavirus_form.groups") { locale_form_groups_fixture }
+    end
+
+    it "returns an array of result link rows" do
+      expect(results_rows).to eq(results_row_fixture)
```
danwilliams, line 108: Is `match_array` a better fit here, or is ordering important?

spec/lib/content_exporter_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+
+    it "returns an empty string value when a link doesn't have :show_to_nations criteria" do
+      expect(results_rows[0][:show_to_nations]).to eq("")
+    end
+
+    it "returns a single string when a link has a single :show_to_nations criteira" do
+      expect(results_rows[2][:show_to_nations]).to eq("Wales")
+    end
+
+    it "returns a singe string with joined with OR when a link has multiple :show_to_nations criteira" do
```
danwilliams, line 119: Here `single` is misspelled as `singe`.

spec/lib/content_exporter_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+      expect(results_rows[0][:text]).to eq("This is a row that only has text, it will appaer like a paragraph")
+    end
+
+    it "returns a boolean support_and_advice true when the link is under support_and_advice" do
+      expect(results_rows[3][:support_and_advice]).to be(true)
+    end
+
+    it "returns a boolean support_and_advice false when the link is under items" do
+      expect(results_rows[0][:support_and_advice]).to be(false)
+    end
```
danwilliams, line 149: These tests are redundant given the test on line 42, where the equality of the received array is checked.

The tests, or rather the test descriptions on these lines, do add context however. As functionally they are commentary; might it make more sense to just make them be comments on/in the fixture definition, stating why each line is included?

(Note - there isn't a strong case for change here; overly verbose tests aren't really a problem, and I'm happy to be pushed back on this!)

spec/lib/content_exporter_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+  },
+  group_two: {
+    subgroup_one: {
+      title: "I am the title for Group two subgroup one",
+      items: [{ id: "0006", text: "This is a row that only has text, it will appaer like a paragraph" }],
+    },
+  },
+}
+
+locale_form_groups_fixture = {
```
danwilliams, line 48: It'd be more maintainable to convert the results row fixture to a CSV and use that as the `csv_fixture`. We're not testing the 'csv' module here, so it's safe to use.

spec/lib/content_exporter_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+      "0005,I am the title for Group one,I am the title for Group one subgroup two,false,\"This is a row that only has text, it will appaer like a paragraph\",\"\",\"\"\n" \
+      "0006,I am the title for Group two,I am the title for Group two subgroup one,false,\"This is a row that only has text, it will appaer like a paragraph\",\"\",\"\"\n"
+    let(:csv) { ContentExporter.generate_results_link_csv }
+
+    before do
+      allow(I18n).to receive(:t).with("results_link") { locale_results_link_fixture }
+      allow(I18n).to receive(:t).with("coronavirus_form.groups") { locale_form_groups_fixture }
+    end
+
+    it "outputs a csv with correct headers" do
```
danwilliams, line 191: This test is redundant given the test @195.

spec/lib/tasks/content_spec.rb
```diff
@@ -0,0 +1,17 @@
+RSpec.describe "content:export_results_to_csv" do
+  include_context "rake"
+
+  before do
+    allow(ContentExporter).to receive(:generate_results_link_csv).and_return("csv\n woop")
+  end
+
+  it "should include environment as a prerequsite" do
```
danwilliams, line 8: Here `prerequisite` is misspelt as `prerequsite`.

lib/content_exporter.rb
```diff
@@ -0,0 +1,44 @@
+require "csv"
+
+class ContentExporter
```
danwilliams, line 3: This is a class with only static methods, it's probably better to use a module with module functions.

spec/lib/content_exporter_spec.rb
huwd: :man_facepalming: thanks for catching that!

spec/lib/content_exporter_spec.rb
huwd: Ah that cool, tbh i hadn't considered `match_array` good shout.
Humm i think ordering is significant though, we render things in the order they appear in the yaml, and i think content designers will want the sheet to be the "order rendered" too.

So i think we should check for ordered equality here.

spec/lib/content_exporter_spec.rb
huwd: Yeah fair comment, I guess i'm worried that there's a lot going on there and wanted to follow a "Tests as documentation" method.

I was thinking, downside is probably that redundant tests slow down test suites and that slows our deploys... but the extra time here seems tiny. And i suppose you'd have to update two expecations instead of one if we made a change...

Humm, i suppose as someone coming to this afresh, did having them there help you understand what I was trying to do? If so i'd rather keep em. If not i'll remove them?

Comments could work too, but if there's minimal cost to having these i might leave them as they are for now.

spec/lib/tasks/content_spec.rb
huwd: ... :man_facepalming: thank you...

lib/content_exporter.rb
huwd: What are the advantages to module functions over a class with static methods?
I'm not particularly attached to this way of doing things, just seemed like a convenient hack.

What do module functions get me instead?

spec/lib/content_exporter_spec.rb
huwd: Yeah fair,
I think because Rspec does a string match here you basically get a massive unreadable blob of red next to a massive unreadable green blog.

When mucking around with the fixtures I found those tests qutie difficult to parse if there was a failure. This one at least had a separate row I could check to see if adding a new column had been added quickly.

But perhaps that's just me!

spec/lib/content_exporter_spec.rb
huwd: Do you mean take this and turn like @177 into something like:
```
let(:csv_fixture) { CSV.generate { |csv| csv << locale_form_groups_fixture} }
```

Or something similar?
So one fixture instead of two?

lib/content_exporter.rb
huwd: :thinking: thinking about this humm... i guess a class has very little meaning if you can't make an instance right?
So perhaps that's why i should use `module` here instead...

huwd: Humm had a read through this: https://dvelp.co.uk/articles/ruby-class-vs-module
I think it's making sense now:
> As a rule of thumb, if you create a Class and it has no instance methods,
you shouldn't be using a Class

Will switch to module now

huwd: Done here: https://github.com/alphagov/govuk-coronavirus-find-support/pull/216/commits/305ed15f2f834d9f2a738c3923d42c7916eada0b

spec/lib/content_exporter_spec.rb
huwd: Humm played with a few options here,
Couldn't find something that made sense. Perhaps we can merge this in and follow up with a few maintainance tweaks in a PR once we've got content updating from these exports / imports?

spec/lib/content_exporter_spec.rb
danwilliams: I'm quite on the fence tbh - they certainly do no harm! 🙂

spec/lib/content_exporter_spec.rb
danwilliams: 👍 Understood, that does make sense!

spec/lib/content_exporter_spec.rb
danwilliams: I was going for the one fixture over two, but looking into implementing it in more detail, the code overlap with the method is too large to retain the tests validity. 

It's probably best left as is 🙂

PR #228, opened 2020-05-20
Altered deploy to account for staging and production metrics routes

danwilliams:
Very minor change to fix the route mappings, which applied production routes to staging, in the same manner as https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/324

Presumably the same note applies here too:
> We should manually fix the issue in staging prior to merging.


bilbof:
@danwilliams looks good, please could you rebase and merge?

PR #238, opened 2020-05-22
(2/2) Conditional Support + Advice Section in results page.

steveholmes:
What
----
The results page needs to distinguish between different types of links and offer users clearer expectations of onwards journeys.

Add markup with a condition for "Support and Advice" sections within a group.

For a given group, if there are "Support and Advice" links, then render the heading + any links. If there are no links then do not show the heading.

NB: A "group" may not have any Support + Advice links. But if there are Support and advice links there will always be "main" links directly under the `h3`

How
---

Support and advice items are listed in the en.yml file and are handled in the same way as regular items. The support items will appear below the regular items.

How to review
-------------

In development, manual testing can be done by adding `support_and_advice_items` at the same level as `items` in the results section of the en.yml file.

In staging or production, testing will need to be coordinated with the cards
https://trello.com/c/Am4icV9a/399-results-page-update-1-3-change-heading-sizes
https://trello.com/c/rbRyihX0/401-results-page-update-3-3-rearrange-content

Links
-----

https://trello.com/c/RWmAHtdt/400-results-page-update-2-3-conditional-support-advice-section

### DIFF
--- a/app/views/components/_actions-group.html.erb
+++ b/app/views/components/_actions-group.html.erb
@@ -14,7 +14,7 @@
           <% %i[items support_and_advice_items].each do |item_type| %>
             <% if subsection[item_type]&.any? %>
               <% if item_type == :support_and_advice_items %>
-                <h3 class="govuk-heading-s"><%= t("coronavirus_form.results.support_and_advice_text") %></h3>
+                <h4 class="govuk-heading-s"><%= t("coronavirus_form.results.support_and_advice_text") %></h4>
               <% end %>
               <ul class="govuk-list">
               <% subsection[item_type].each do |item| %>
--- a/app/views/components/_actions-group.html.erb
+++ b/app/views/components/_actions-group.html.erb
@@ -11,9 +11,13 @@
       <% subsections.each do |subsection| %>
         <div data-ec-list-subsection="<%= subsection[:title] %>">
           <h3 class="govuk-heading-m"><%= subsection[:title] %></h3>
-          <% if subsection[:items].any? %>
-            <ul class="govuk-list">
-              <% subsection[:items].each do |item| %>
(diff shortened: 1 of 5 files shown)


app/views/components/_actions-group.html.erb
```diff
@@ -11,9 +11,13 @@
       <% subsections.each do |subsection| %>
         <div data-ec-list-subsection="<%= subsection[:title] %>">
           <h3 class="govuk-heading-m"><%= subsection[:title] %></h3>
-          <% if subsection[:items].any? %>
-            <ul class="govuk-list">
-              <% subsection[:items].each do |item| %>
+          <% %i[items support_and_advice_items].each do |item_type| %>
+            <% if subsection[item_type].any? %>
+              <% if item_type == :support_items %>
+                <h3 class="govuk-heading-s"><%= t("coronavirus_form.results.support_and_advice_text") %></h3>
```
injms, line 17: At first glance I thought the the `h3` on line 17 should be an `h4` since visually this heading appears to be nested underneath the subsection title on line 13.

But looking at the Trello card it could semantically make sense to be either a `h3` or a `h4`.

@alex-ju - what do you think?

alex-ju: I'm getting an `undefined method `any?' for nil:NilClass` error while trying to preview the component (e.g. `http://localhost:5000/component-guide/actions-group`), we may need to initialise `subsection ` as an array or so. I'll go through the Trello cards to try and understand the requirements.

alex-ju: Thanks for fixing this @steveholmes. To me, it seems like an `h4` and since the component makes sure that an `h3` exists and we're not skipping levels I would suggest we go with that.

PR #251, opened 2020-05-28
Create a Baseline from from the AlphaGov Rep (2020-05-28)

rickclare:
1. Create a Baseline of the GovWales version, from the AlphaGov commit https://github.com/alphagov/govuk-coronavirus-find-support/commit/6c39ee4a57b0e97b7792e9163c4c2b3e1eb947a7

2. Remove the first "Nations" question

3. Apply Welsh specific en.yml content


rickclare:
PR opened in error

PR #252, opened 2020-05-28
Export data for number of times checkboxes on need-help-with are selected per day

Rosa-Fox:
What
----
- Export data for number of times checkboxes on `/need-help-with` are selected per day.

Describe what you have changed and why.

- Decided with Business Analyst to keep this separate to the [main data-export](https://github.com/alphagov/govuk-coronavirus-find-support/blob/master/app/controllers/coronavirus_form/data_export_controller.rb). This makes reporting easier for her in Datastudio. The main data-export reports on the count per combination of checkboxes selected, rather than on individual selections.

- [Automating daily reporting to Google sheets](https://github.com/alphagov/govuk-coronavirus-find-support/blob/master/docs/result_links_imports.md) will happen in a follow on task.

How to review
-------------

Describe the steps required to review and test the changes.

- Export data locally by visiting the following end point: http://localhost:5000/data-export-checkbox.csv

- Run the tests.

- Any refactoring ideas appreciated!

Links
-----

[Trello](https://trello.com/c/C0F7RP6P/260-spike-on-analytics-for-question-with-checkboxes)

### DIFF
--- a/app/controllers/coronavirus_form/data_export_checkbox_controller.rb
+++ b/app/controllers/coronavirus_form/data_export_checkbox_controller.rb
@@ -1,0 +1,43 @@
+require "csv"
+
+class CoronavirusForm::DataExportCheckboxController < ApplicationController
+  include DataExportCheckboxHelper
+
+  if ENV.key?("DATA_EXPORT_BASIC_AUTH_USERNAME") && ENV.key?("DATA_EXPORT_BASIC_AUTH_PASSWORD")
+    http_basic_authenticate_with(
+      name: ENV.fetch("DATA_EXPORT_BASIC_AUTH_USERNAME"),
+      password: ENV.fetch("DATA_EXPORT_BASIC_AUTH_PASSWORD"),
+    )
+  end
+
+  def show
+    respond_to do |format|
+      format.html
+      format.csv do
+        render csv: produce_csv(usage_statistics(params[:start_date], params[:end_date])),
+               filename: "data-export"
+      end
+    end
+  end
+
+private
+
+  def produce_csv(results)
+    question = I18n.t("coronavirus_form.groups")[:filter_questions][:questions][:need_help_with][:title]
+    csv_data = CSV.generate(col_sep: "|") do |csv|
+      csv << %w[question answer date count]
+      results.each do |_question, value|
+        value.each do |k|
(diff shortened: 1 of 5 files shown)


app/controllers/coronavirus_form/data_export_checkbox_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+               filename: "data-export"
+      end
+    end
+  end
+
+  def checkbox_options
+    options = I18n.t("coronavirus_form.groups")[:filter_questions][:questions][:need_help_with][:options]
+
+    options.index_by do |_option|
+      :need_help_with
```
leenagupte, line 25: In the form, this value is saved as "#{option.parameterize.underscore}". Should the same value be used here?

app/helpers/data_export_checkbox_helper.rb
```diff
@@ -0,0 +1,47 @@
...
+        date: created_date,
+        count: count,
+      }],
+    }
+  end
+
+  def responses(start_date, end_date)
+    FormResponse
+    .where(created_at: start_date..end_date)
+    .select(Arel.sql("created_at::date, form_response -> 'need_help_with'"))
```
leenagupte, line 39: In my travels I discovered you don't need this `select`. The `pluck` works by itself. 😁

app/helpers/data_export_checkbox_helper.rb
```diff
@@ -0,0 +1,47 @@
...
+    end_date = end_date ? Date.parse(sanitize(end_date)).end_of_day : Time.zone.today.end_of_day
+
+    results = {}
+    counts = Hash.new(0)
+
+    responses(start_date, end_date).each do |response|
+      counts[response] += 1
+    end
+
+    counts.map do |(option_text, created_date), count|
```
leenagupte, line 13: Teensy nitpick (as I'm the one that led you astray): could use an `each` here.

Rosa-Fox: 👍

app/helpers/data_export_checkbox_helper.rb
Rosa-Fox: 🎉

PR #258, opened 2020-06-01
Update escape link component

alex-ju:
What
----

Open escape link page in a new tab and replace the original page to prevent the perpetrator from easily navigating backward. 

How to review
-------------

Deploy the app locally and click the 'escape link' on any page.

Links
-----

[Trello cards](https://trello.com/c/dwW7wfx3)

### DIFF
--- a/spec/javascripts/components/escape-link.spec.js
+++ b/spec/javascripts/components/escape-link.spec.js
@@ -1,0 +1,35 @@
+/* eslint-env jasmine, jquery,  EscapeLink, */
+
+describe('Escape link component', function () {
+  'use strict'
+
+  var container
+  var escapeLinkElement
+  var escapeLinkModule
+
+  beforeEach(function () {
+    container = document.createElement('div')
+    container.innerHTML = '<a class="app-c-escape-link govuk-link" rel="nofollow noreferrer noopener" target="_blank" data-module="app-escape-link" href="https://www.gov.uk/">Leave this site</a>'
+    document.body.appendChild(container)
+    escapeLinkElement = document.querySelector('[data-module="app-escape-link"]')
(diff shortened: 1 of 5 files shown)


spec/javascripts/components/escape-link.spec.js
```diff
@@ -0,0 +1,35 @@
...
+  var escapeLinkElement
+  var escapeLinkModule
+
+  beforeEach(function () {
+    container = document.createElement('div')
+    container.innerHTML = '<a class="app-c-escape-link govuk-link" rel="nofollow noreferrer noopener" target="_blank" data-module="app-escape-link" href="https://www.gov.uk/">Leave this site</a>'
+    document.body.appendChild(container)
+    escapeLinkElement = document.querySelector('[data-module="app-escape-link"]')
+    escapeLinkModule = new EscapeLink(escapeLinkElement)
+    escapeLinkModule.replaceCurrentPage = function () {}
```
alex-ju, line 16: Jasmine crashes on any `window.location.replace` calls, so I had to mock this in tests to avoid it.

alex-ju:
> This seems to work, but it results in two tabs on bbc.co.uk, is that intended?

Well, I don't have other input on this, and using the same URL means the browser will use an already cached page, which makes it quicker.

bilbof:
I wonder whether it'd be possible to open just one tab? If not, I'm happy for this to be merged in :)

alex-ju:
> I wonder whether it'd be possible to open just one tab? I'm happy for this to be merged in regardless :)

Thanks for having the chat around this. To report properly, opening a new tab helps with hiding the browser back link, and because we're not able to close the original page (`window.close` only works for pages opened via JavaScript) we're replacing it with another one (which helps with clearing the last page from history, but unfortunately not all of them).

PR #280, opened 2020-06-08
Handle empty rows in the form.

steveholmes:
Why
---
The google sheet has some pre-filled formulae in the hidden columns of the spreadsheet, ready for new content to be added. If a row contains only empty or '#N/A' we should ignore these rows.

What
----
Ignore rows which contain only blank or '#N/A' error values.

How to review
-------------

Follow instructions here to run an import into local dev environment. There should be no errors during the import.
https://github.com/alphagov/govuk-coronavirus-find-support/blob/master/docs/result_links_imports.md

Links
-----

https://trello.com/c/Tx1P5tbk/559-fix-google-sheet-import-export-process-for-content

### DIFF
--- a/lib/content_importer.rb
+++ b/lib/content_importer.rb
@@ -55,6 +55,8 @@ module_function
   def import_results_links(csv_path)
     csv = CSV.read(csv_path, { headers: true })
     output = csv.each_with_object({}) do |csv_row, results_links|
+      next if blank_row?(csv_row)
+
       support_and_advice_items = csv_row.fetch("support_and_advice")
       group_key = csv_row.fetch("group_key").to_sym
       subgroup_key = csv_row.fetch("subgroup_key").to_sym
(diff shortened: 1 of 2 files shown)


lib/content_importer.rb
```diff
@@ -99,4 +101,11 @@ def overwrite_locale_links(input_locale_path, locale_hash, output_locale_path =
       file.write(existing_locale_file.to_yaml)
     end
   end
+
+  def blank_row?(row)
+    row.each do |_heading, value|
+      return false unless [nil, "#N/A"].include? value
+    end
+    true
```
bilbof, line 109: I think a more idiomatic way to say this would be
```ruby
def blank_row?(row) row.values.any? { |v| [nil, "#N/A"].include?(v) }; end
```

steveholmes: Originally I found there wasn't a `values` method but after searching again I realised `fields` gives us what we need 👍

bilbof: Ah cool, thanks 👍

PR #294, opened 2020-06-11
Rename the living with vulnerable route

JonathanHallam:
A previous commit changed the copy for the "Are you worried about going into work" question. It didn't change the route so that was still `/living-with-vulnerable`, a phrase that wasn't in the copy any more. This commit changes the names of the controller, view and spec for that path, changes the routes and adds a redirect for the old route. This means that the route the user sees now matches the copy on the page.


config/routes.rb
```diff
@@ -22,6 +22,9 @@
     # Redirect for old route (301 is default)
     get "/where-live", to: redirect("/nation")
 
+    # Redirect for old route (301 is default)
+    get "/living-with-vulnerable", to: redirect("worried-about-work")
```
leenagupte, line 26: This is a great idea!

PR #337, opened 2020-07-08
Update pipeline deploy commands to use the latest cli version

leenagupte:
What
----

The pipeline failed when deploying to staging. 

```
'v3-create-app' is not a registered command. See 'cf help -a'
'v3-apply-manifest' is not a registered command. See 'cf help -a'
```

The reason for this is that PaaS changed the Docker image to use a newer version of the CLI.

As a quick fix, we tagged a version that we knew to work with the cli that we are currently using. 

This updates the CLI commands so that we can go back to using `tag: latest` 

https://hub.docker.com/r/governmentpaas/cf-cli/tags

The changes are taken from the [list of updated CLI v7 commands](https://docs.cloudfoundry.org/cf-cli/v7.html#table)

How to review
-------------

NOTE: I'm not sure how to test this locally.
Install version 7 of the CLI:
```
brew install cloudfoundry/tap/cf-cli@7
```

Links
-----

[Trello cards](https://trello.com/c/o0LxKNog)

### DIFF
--- a/concourse/tasks/deploy-to-govuk-paas.yml
+++ b/concourse/tasks/deploy-to-govuk-paas.yml
@@ -39,7 +39,7 @@ run:
       cf auth
       cf t -o "$CF_ORG" -s "$CF_SPACE"
       cf create-app govuk-coronavirus-find-support || true
-      cf v3-apply-manifest -f manifest.yml
+      cf apply-manifest -f manifest.yml
       cf set-env govuk-coronavirus-find-support CF_STARTUP_TIMEOUT "$CF_STARTUP_TIMEOUT"
 
       cf scale -i "$INSTANCES" govuk-coronavirus-find-support
--- a/concourse/tasks/deploy-to-govuk-paas.yml
+++ b/concourse/tasks/deploy-to-govuk-paas.yml
@@ -61,5 +61,5 @@ run:
       cf set-env govuk-coronavirus-find-support AWS_ERROR_PAGES_BUCKET_NAME "$AWS_ERROR_PAGES_BUCKET_NAME"
       cf set-env govuk-coronavirus-find-support AWS_REGION "$AWS_REGION"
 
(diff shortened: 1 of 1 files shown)


concourse/tasks/deploy-to-govuk-paas.yml
```diff
@@ -38,8 +38,8 @@ run:
       cf api "$CF_API"
       cf auth
       cf t -o "$CF_ORG" -s "$CF_SPACE"
-      cf v3-create-app govuk-coronavirus-find-support || true
-      cf v3-apply-manifest -f manifest.yml
+      cf push govuk-coronavirus-find-support --strategy rolling || true
+      cf apply-manifest -f manifest.yml
```
richardTowers, line 42: I think it would be closer to the original to use `cf create-app` (note: dropping the `v3-` prefix), and then to replace `cf v3-zdt-push` with `cf push --strategy rolling`.

As a bit of context, the `cf create-app` and `cf apply-manifest` commands are like little bits of what happens when you run `cf push` (which tries to do everything at once).

Ideally, we wouldn't have all this scripting and we'd just have a single `cf push` command with a manifest that contained everything we needed. I couldn't make that work the way I wanted before - if I remember correctly the issue was that I needed to set a bunch of environment variables to secret values, and `cf v3-zdt-push` wouldn't let me pass them in on the command line, so I had to set them separately with `cf set-env`.

I _think_ now that we're on the newer CLI we could fix a lot of the tech debt here, and just state explicitly in the manifest file which variables we need to provide.

I'd suggest experimenting in a local terminal pushing to a sandbox space would be a good idea though.

leenagupte: Thanks for the advice @richardTowers! I've made the suggested changes and fixed up. Now to find a sandbox to test it...

PR #353, opened 2020-07-16
690 Remove 'Are you able to go out for food, medicine, or health reasons?'

reggieb:
Removed are you able to go out page

Why
----

In the find support service, if a user selects they can't get food, we ask this able_to_go_out question:

Are you able to go out for food, medicine, or health reasons?

If a user selects No, I should not go out because I have a medical condition which means I’m extremely vulnerable to coronavirus then we show them a link to the extremely vulnerable people service.

The extremely vulnerable people service will no longer exist, so we are not showing people this option.

This means this question is not needed at all

What
----
Removed the controller, views, route and specs

Links
-----

[Trello card](https://trello.com/c/SUHEMkvf/690-ideally-by-17-july-remove-are-you-able-to-go-out-for-food-medicine-or-health-reasons)


reggieb:
Thank you @Rosa-Fox I'll work on that on Monday.

PR #354, opened 2020-07-17
Recurring Rake task to report on broken links

steveholmes:
What
----

Take links found in our content and run them against link-checker-api to determine any broken, redirecting, slow or malicious links

We need to be able to identify whether there are any broken links in the service to ensure users aren't being sent to pages with errors, malicious sites or withdrawn guidance.

How to review
-------------

Bearer tokens need to be configured.
`LINK_CHECKER_API_BEARER_TOKEN` in ENV settings

task can be checked locally by setting bearer tokens and running the rake task
`rake link-checker`

Links
-----

https://trello.com/c/JldlpFxD/435-add-recurring-rake-task-to-report-on-broken-links


lib/link_checker.rb
```diff
@@ -0,0 +1,49 @@
+require "uri"
+
+module LinkChecker
```
reggieb, line 3: This really feels like it should be a class rather than a module. Using a module means that the Rake task using this functionality has to know a lot about how it works - the logic is split between the rake task and the module, because the module methods are effective chunks of functional code.

I think it would be better to initiate an instance, passing in a hash of strings (output from a `YAML.safe_load`) on initialization, and then output a report.

lib/link_checker.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+    end
+    urls
+  end
+
+  def get_urls_from_locale_files
+    # For each file, gather the urls as a set and then combine the sets together using reduce
+    Dir.glob("config/locales/*.yml")
+     .map { |locale_filename| gather_urls YAML.safe_load(File.read(locale_filename)) }
+     .reduce(:+)
```
reggieb, line 28: Combining the gathering of data from YAML files with extracting the URLs from the strings in one class level method, make this very difficult to test. Hence the spec for this method is of full stubbing. IMO This should be broken down into two methods. One to gather the raw data from the YAML files, and another to extract the urls from that data. Then each part of the process can be tested in isolation and without stubbing.

concourse/tasks/run-link-checker.yml
```diff
@@ -0,0 +1,29 @@
...
+  source:
+    repository: governmentpaas/cf-cli
+    tag: latest
+params:
+  CF_API: https://api.cloud.service.gov.uk
+  CF_USERNAME: ((paas-username))
+  CF_PASSWORD: ((paas-password))
+  CF_ORG: govuk_development
+  SENTRY_DSN: https://((sentry-dsn))
+  PUBLISHING_API_BEARER_TOKEN:
```
Rosa-Fox, line 13: Before we merge, I will need to add the bearer tokens to secrets.. this line will need to become:

```
PUBLISHING_API_BEARER_TOKEN: ((publishing-api-bearer-token))
```

The `(())` syntax means that it is being read in from GOV.UK secrets.

concourse/tasks/run-link-checker.yml
```diff
@@ -0,0 +1,29 @@
...
+  LINK_CHECKER_API_BEARER_TOKEN:
```
Rosa-Fox, line 14: Same as the above comment. This will need to be:

```
LINK_CHECKER_API_BEARER_TOKEN: ((link-checker-api-bearer-token))
```

concourse/tasks/run-link-checker.yml
```diff
@@ -0,0 +1,29 @@
...
+    - '-c'
+    - |
+      set -eu
+
+      cf api "$CF_API"
+      cf auth
+      cf t -o "$CF_ORG" -s "$CF_SPACE"
+      cf set-env govuk-coronavirus-find-support PUBLISHING_API_BEARER_TOKEN "$PUBLISHING_API_BEARER_TOKEN"
+      cf set-env govuk-coronavirus-find-support  LINK_CHECKER_API_BEARER_TOKEN "$LINK_CHECKER_API_BEARER_TOKEN"
+      cf run-task govuk-coronavirus-find-support "rake link_checker"
```
Rosa-Fox, line 29: `cf run-task govuk-coronavirus-find-support "bundle exec rake link_checker"`

reggieb:
The change to `LinkChecker` look good (IMO) and have simplified the specs which is good. However, I think Rosa's comments still need addressing

Rosa-Fox:
@steveholmes looks good! Like Rob said, you need to make the changes I suggested, rebase and then it's good to go. I ran it locally, is cool :)

steveholmes:
Thank you for the help, should be good now :+1:

PR #341, opened 2020-07-10
Allow show_to_nations to be [] or ""

Rosa-Fox:
What
----

Describe what you have changed and why.

- `show_to_nations` can now be in array or string format.
- The reason for this change is that is easy to format the YAML incorrectly for single entries.
- It should prevent us from getting [these errors](https://sentry.io/organizations/govuk/issues/1772131367/?project=5192076&referrer=slack).

Co-authored-by: Rob Nichols @reggieb <rob.nichols@digital.cabinet-office.gov.uk>

How to review
-------------

:ship: Since this service is continuously deployed, all testing must be done
before the pull request is merged. :ship:

- Review the code
- Run the tests

Links
-----

[Trello](https://trello.com/c/yAcNAwXN/677-investigate-results-link-data-export-error)

PR #357, opened 2020-07-20
Remove the google sheet importer.

steveholmes:
What
----
We want to change the way we import content into the results page, from a Google Sheets to uploading directly onto the YAML file.

Why
---
The Google Sheets method was built in case people didn't know how to use YAML. But it's a bit fiddly to work with and the team know how to use YAML, so let's make it easier for the team.

Agreed to keep the Google sheet exporter, so that Content Designers have a cleaner way of viewing the results. Remove only the importer for now.

How to review
-------------

Nothing to run, visual check that the importer has been removed.

Links
-----

https://trello.com/c/I0TkL1mK/698-change-content-uploading-method-from-google-sheets-to-yaml


docs/result_links_imports.md
```diff
@@ -27,20 +25,20 @@ GOOGLE_SHEET_ID="a-google-sheet-id"
 3. Run this take task:
 
 ```
-bundle exec rake content:import_locale_links_from_google_sheet
+bundle exec rake content:extract_results_links
```
Rosa-Fox, line 28: I think this should be:

```
 bundle exec rake content:export_results_to_csv
```

However this task now fails as we allow strings in the YAML as a result of [this PR](https://github.com/alphagov/govuk-coronavirus-find-support/pull/341) :*( 

Should we just change the rake command here and then make fixing the export a separate Trello task?

docs/result_links_imports.md
```diff
@@ -27,20 +25,20 @@ GOOGLE_SHEET_ID="a-google-sheet-id"
...
+The google sheet should import results regularly, but there is a option in the sheet menu `Find Support Service -> Refresh Data` to import the data ad-hoc.
 
-You can also manually import from a provided CSV file
+The google sheet will access the service using the url
+`https://find-coronavirus-support.service.gov.uk/data-export-results-links.csv`
 
+There is also a rake task allowing a developer to run locally without pushing the file to google. 
 ```
-bundle exec rake content:import_locale_links[file_path]
+bundle exec rake content:extract_results_links
```
Rosa-Fox, line 43: See comment above!

docs/result_links_imports.md
steveholmes: Ah good spot, thx, I'll update

steveholmes: I've rebased for the earlier YAML fix and corrected the bundle command in both places.

PR #363, opened 2020-07-21
687 Put feedback call to action under each section

reggieb:
What
----
- Moved feedback to dedicated section of translation file
- Added feedback title entries for each question group in file
- Moved feedback output into dedicated partial
- Rendered feedback partial within action group
- Made title labels consistent (so they are not renamed as headings
  and then renamed back to titles again)
- Added controller test so fine grained behaviour could be tested
- Prevent the hash generated by `result_groups(session)` being created three times 
  when only needed once in results view

Why
----
The triage tool doesn't get a lot of feedback - but the feedback it does get is very skewed towards users that don't get any results (and therefore find the service 'useless'). This could be because users don't see the feedback CTA unless they don't get any results. We should explore making it more visible to all users so we get a better reflection of how our users are finding the tool.

Example of change
----

mulitple_feedback_callouts

Links
-----

[Trello card](https://trello.com/c/2vze3A5S/687-put-feedback-call-to-action-under-each-section)

### DIFF
--- a/app/views/coronavirus_form/results.html.erb
+++ b/app/views/coronavirus_form/results.html.erb
@@ -1,4 +1,5 @@
-<% title = results_title(result_groups(session)) %>
+<% session_result_groups = result_groups(session) %>
+<% title = results_title(session_result_groups) %>
 
 <% content_for :title do %>
   <%= title %>
@@ -28,24 +29,16 @@
   %>
 <% end %>
 
-<% if result_groups(session).empty? %>
+<% if session_result_groups.empty? %>
   <%= sanitize(t("coronavirus_form.results.no_results")) %>
   <%= sanitize(t("coronavirus_form.results.no_results_#{session[:nation].downcase.gsub(" ", "_")}")) %>
+  <%= render "components/feedback" %>
 <% else %>
   <div data-track-ec-list="Find coronavirus support">
-    <% result_groups(session).each do |group| %>
-      <%= render "components/actions-group", {
-        title: group[1][:heading],
-        subsections: group[1][:questions]
-      } %>
+    <% session_result_groups.values.each do |group| %>
+      <%= render "components/actions-group", group %>
     <% end %>
   </div>
-<% end %>
-
-<%= render "components/callout", {
-  title: "Help us improve"
-} do %>
-  <%= link_to t("coronavirus_form.results.feedback.link_text"), t("coronavirus_form.results.feedback.link_href"), class: "govuk-link" %>
 <% end %>
 
 <% content_for :escape_link do %>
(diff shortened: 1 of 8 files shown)


app/views/coronavirus_form/results.html.erb
```diff
@@ -28,26 +29,18 @@
...
   <%= sanitize(t("coronavirus_form.results.no_results_#{session[:nation].downcase.gsub(" ", "_")}")) %>
+  <%= render "components/feedback" %>
 <% else %>
   <div data-track-ec-list="Find coronavirus support">
-    <% result_groups(session).each do |group| %>
-      <%= render "components/actions-group", {
-        title: group[1][:heading],
-        subsections: group[1][:questions]
-      } %>
+    <% session_result_groups.values.each do |group| %>
```
JonathanHallam, line 38: nice refactor 🎉

config/locales/en.yml
```diff
@@ -1139,3 +1135,15 @@ en:
           text: Advice on supporting children’s mental health during coronavirus (Young
             Minds)
           href: https://youngminds.org.uk/find-help/for-parents/supporting-your-child-during-the-coronavirus-pandemic/
+  feedback:
+    link_text: Give feedback on this service
+    link_href: https://www.gov.uk/done/find-coronavirus-support
+    titles:
+      feeling_unsafe: Was this information on what to do if you’re feeling unsafe or worried about someone else helpful?
```
JonathanHallam, line 1142: I wonder if there's a way to do this more programmatically. Not a blocker for this by any means but given that these are so generic / tied to the title of the page we might be able to cut down on this slightly at some point

spec/controllers/coronavirus_form/results_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    before do
+      session[:nation] = "England"
+    end
+
+    it "shows default feedback" do
+      expect(subject.body).to include(I18n.t("feedback.titles.default"))
+    end
+
+    context "when an answer is present" do
```
JonathanHallam, line 19: nice :)

config/locales/en.yml
reggieb: I did consider doing some sort of programmatic text construction as they all currently have the structure `"Was this information on #{page_specific_content} helpful?"`. However, I thought it very likely that the text would change and the cleverer the solution, the more difficult it would be to change the text structure. So in the end I went for the simple solution.

JonathanHallam: Yeah, that makes sense, I think I would have ended up doing what you have here. Cards on the table I've always really disliked this file because of how long it is. Probably just me getting a bee in my bonnet

reggieb: Oh! I was very tempted to split this feedback segment out into it's own translation file. It just seemed out of keeping with the rest of the way the translations are presented. Breaking the file down into it three or four main segments would make a lot of sense to me, but I think that should be done in a separate PR. It might even be nice to have a translation file for each question - but that would require a big refactor.

PR #366, opened 2020-07-24
Move nation question to the end of the flow

steveholmes:
What
----

The question regarding which nation has been moved to the final question.

This question was the first question in the flow.

- Changed logic in the application controller that sets the def first_question_path...  to  /need-help-with path.
- We also checked the session exists (def check_first_question_answered) via checking the first path, so was updated.
- When the final question is submitted, we persisted the answers to the database. This previously had some complexity as the last question would change depending on which checkboxes were selected.
- Given that the /nation question should be shown to all users, we can go back to only persisting the data in that controller... which meant removing from all of the other places it was added to

How to review
-------------
Ensure the flow has been correctly implemented, including back links.

Links
-----

https://trello.com/c/57lE8k8g/694-move-the-where-do-you-want-information-about-question-to-the-end-of-the-flow


Gemfile.lock
```diff
@@ -482,4 +482,4 @@ RUBY VERSION
    ruby 2.6.6p146
 
 BUNDLED WITH
-   1.17.2
+   1.17.3
```
leenagupte, line 485: ❓ I'm not sure what this is doing here.

app/controllers/application_controller.rb
```diff
@@ -56,15 +56,9 @@ def session_expired
     redirect_to session_expired_url
   end
 
-  def check_first_question_answered
```
leenagupte, line 59: Could you add a note to the commit message about why this is being removed? Something about the filter question already being checked etc.

app/controllers/coronavirus_form/need_help_with_controller.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class CoronavirusForm::NeedHelpWithController < ApplicationController
-  before_action :check_first_question_answered
+  # before_action :check_first_question_answered
```
leenagupte, line 4:
```suggestion
```

This line should be deleted rather than commented out.

app/helpers/questions_helper.rb
```diff
@@ -43,8 +43,8 @@ def first_question
     questions_to_ask.first
   end
 
-  def first_question_seen?
-    session[:questions_to_ask].present?
+  def last_question_seen?
+    session[:nation].present?
```
leenagupte: Can you add a note about why this check is being changed?
Also, `first_question_seen?` is still being stubbed [here](https://github.com/alphagov/govuk-coronavirus-find-support/blob/master/spec/requests/need_help_with_spec.rb#L4)

Gemfile.lock
steveholmes: :+1:

app/controllers/coronavirus_form/need_help_with_controller.rb
steveholmes: :+1:

app/controllers/application_controller.rb
steveholmes: :+1:

app/helpers/questions_helper.rb
steveholmes: :+1:

steveholmes:
@leenagupte Coverage is falling because I deleted some tested lines.
Previous the untested methods were, say for example, 1 out 10 lines (10%). If I delete 5 tested lines, then the untested line is now 1 out 5 so the file is now 20% untested :(
That's what caused my comment in slack about removing the untested method `set_session_history`

PR #370, opened 2020-07-28
Fixes analytics on results page

DilwoarH:
The Analytics code was looking for `data-ec-list-subsection`, the use of `data-ec-list-question` is not documented anywhere.

This was caused by this commit: https://github.com/alphagov/govuk-coronavirus-find-support/commit/a80fab8721786ab27b97be063a32f86619716d8c#diff-d4e7ecc147ff9ecf21ade9419e00a274R12

https://trello.com/c/Fe34Ewsa/726-ga-tracking-stopped-on-results-page

### DIFF
--- a/app/views/components/_actions-group.html.erb
+++ b/app/views/components/_actions-group.html.erb
@@ -9,7 +9,7 @@
     </div>
     <div class="govuk-grid-column-two-thirds">
       <% questions.each do |question| %>
-        <div data-ec-list-question="<%= question[:title] %>">
+        <div data-ec-list-subsection="<%= question[:title] %>">
           <h3 class="govuk-heading-m"><%= question[:title] %></h3>
           <% %i[items support_and_advice_items].each do |item_type| %>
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/analytics-enhancedEcommerce.js
```diff
@@ -7,7 +7,13 @@ var enhancedEcommerceTracking = function (d) {
...
+    var present = el.parentElement
+    
+    if (present) {
+      present = present.querySelectorAll(selector)
+    } else {
+      return null 
+    }
 
     if (present.length < 1) {
       return ancestor(el.parentElement, selector)
```
injms, line 19: Just noticed on line 42:

```javascript
var subsection = ancestor(listItem, '[data-ec-list-subsection]').getAttribute('data-ec-list-subsection')
```

will need a check to make sure that `.getAttribute()` isn't used when `null` is returned from `ancestor()`.

PR #385, opened 2020-08-11
Freelance, self employed and out of work amends

hannahjump:
https://trello.com/c/t3VJ1Wbh/428-help-users-who-are-freelance-contractors-agency-workers

What
----

Changes:
-  include word 'freelancer' with self employed 
- change to being redundant or 'out of work' in all mentions, rather than unemployed
- added universal credit link

Describe what you have changed and why.

How to review
-------------

Describe the steps required to review and test the changes.

Links
-----

Add links to any relevant [Trello cards](https://trello.com/b/9Pvq5z2n/govuk-coronavirus-services-team-doing-board),
Zendesk tickets or Sentry issues.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -271,7 +271,7 @@ en:
         title: Being made redundant or unemployed, or not having any work
         questions:
           self_employed:
-            title: Are you self-employed or a sole trader?
+            title: Are you self-employed, a freelancer, or a sole trader?
             title_caption: Being unemployed or not having any work
             options:
               option_yes:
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -703,6 +703,9 @@ en:
         - id: '0048'
           text: What to do if you’ve been furloughed
           href: https://www.gov.uk/guidance/coronavirus-covid-19-what-to-do-if-youre-employed-and-cannot-work
+        - id: '0048'
```
reggieb, line 706: Having the same id as the previous element look suspicious, but have check with Hanna and through the code, and the ids support a feature that is now redundant - so just a place holder.

However, I'd think it better that a unique number is used, as duplicating the number from the previous element suggests that there is a special relationship between this elements share this id. I'd suggest something obviously different. Like `'1049'`

PR #398, opened 2020-09-07
Move "Are you off work ill?" question

theseanything:
What
----

We are moving the "Are you off work ill?" question from the unemployed section to the going into work section. Also renaming the `Going in to work` option to `Being worried about working, or off work because you’re self-isolating` to make it easier for user's to find advice.

How to review
-------------

Selecting the "Being made redundant or unemployed, or not having any work" and the "Being worried about working, or off work because you’re self-isolating" option from the "What you need help with?" question.

You should be routed to the following question in this order:
- Are you self-employed, a freelancer, or a sole trader?
- Have you been made redundant or told to stop working? (_Having answered no to previous question_)
- Are you worried about going in to work?
- Are you off work because you’re ill or self-isolating?

Answering yes to the "Are you off work because you’re ill or self-isolating?" should show the relevant section on the results page under "Being worried about working, or off work because you’re self-isolating" group.

Links
-----

https://trello.com/c/F7VZa3kL/476-move-question-for-if-youre-off-work-ill-or-self-isolating

### DIFF
--- a/app/controllers/coronavirus_form/are_you_off_work_ill_controller.rb
+++ b/app/controllers/coronavirus_form/are_you_off_work_ill_controller.rb
@@ -36,6 +36,6 @@ private
   end
 
   def group
-    "being_unemployed"
+    "going_in_to_work"
   end
 end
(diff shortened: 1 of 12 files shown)


gclssvglx (approved):
Nice one - looks good 👍

PR #410, opened 2020-09-28
Add route that redirects all GET requests to new app url if ENV present

reggieb:
Reroutes all GET requests to https://www.gov.uk/find-coronavirus-support if app is in production mode

[Trello Ticket](https://trello.com/c/6Pf4DHwl/379-redirect-existing-triage-tool-to-smart-answer)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,6 +6,10 @@ Rails.application.routes.draw do
 
   get "/", to: redirect("https://www.gov.uk/find-coronavirus-support")
 
+  if Rails.env.production?
+    get "*everything", to: redirect("https://www.gov.uk/find-coronavirus-support")
+  end
+
   scope module: "coronavirus_form" do
     first_question = "/need-help-with"
 


config/routes.rb
```diff
@@ -6,6 +6,10 @@
 
   get "/", to: redirect("https://www.gov.uk/find-coronavirus-support")
 
+  if ActiveModel::Type::Boolean.new.cast(ENV["REROUTE_TO_LIVE"])
+    get "*everything", to: redirect("https://www.gov.uk/find-coronavirus-support")
+  end
```
theseanything, line 11: I don't think we need the flag - we'll just deploy redirect after the migration. 

We can also remove all the other routes and just have a single wildcard redirect.

theseanything (changes requested):

config/routes.rb
reggieb: That will break all the tests which will mean the code won't get through CI.

gclssvglx (commented):
Having a flag makes sense - it gives us a quick/easy way to turn the flow back on if anything goes wrong and we need to revert.

theseanything (approved):

theseanything (approved):

